PR #14, opened 2016-07-03
Interactive model editing actions and behaviours. 

tinevez:
This PR adds a core of actions and behaviours that allow for the manual editing and curation of a model. It's the leading branch in current network, and is already merged with `resize-spot-action` and `model-edit-actions`.

The next steps will require the refactoring on how a model is modified in a view.


tinevez:
The finder-like and midnight-commander-like navigation behaviour should go in one class, and the UI should be able to switch from one to another.

PR #22, opened 2016-09-16
Search vertex label

tpietzsch:
opening this PR to have a place for discussion...


tpietzsch:
@tinevez This is almost ready.

My last requests:

The `TrackSchemePanel.SearchAction` conflates iteration of the graph with the iteration. The nested iteration over roots and breadth-first per root should be pulled out into a separate Iterator class.

Also, while the iteration of roots follows `LexicographicalVertexOrder`, the breadth-first does not. Could you please make (and use) a variant of the generic BreadthFirstIterator that is constructed with a `Comparator<V>` for ordering children.`LexicographicalVertexOrder` is not a `Comparator<TrackSchemeVertex>` yet, but that should be straightforward.

tinevez:
- [ ] The nested iteration over roots and breadth-first per root should be pulled out into a separate Iterator class.
- [ ] Make (and use) a variant of the generic BreadthFirstIterator that is constructed with a Comparator<V> for ordering children.
- [ ] Make LexicographicalVertexOrder a Comparator<TrackSchemeVertex>.

tinevez:
Actually I just realized that points 1 & 2 require more work than I thought and wanted to check with you this first.

The BreadthFirstIterator you describe would iterate over the whole graph, regardless of the connected component. Once it reaches the end of the vertices _below_ a root, it would jump to the next root, and do breadth first again. 

All would be fine with my dumb iterator if we had graphs that are only made of trees. But as soon as we have a Y-shaped track we are doomed. 

The Y-shaped track has two roots for a single connected-component. My strategy would iterate over some vertices twice, which breaks the contract of a graph iterator. 

Whole-graph traversal is probably worth some work. Using the roots to jump from a connected component to another is desirable, but we have roots only for TrackSchemeGraph. 
Do you think we can generalize this?

tpietzsch:
@tinevez In principle this would go away if the iterator acts as if there was one virtual "super-root" that had all roots as children. I'm not familiar enough with the graph iterators to judge whether this would be easy for the algorithm to? Could be as easy as pre-load the queue with all roots?

tinevez:
I will work on this.
The idea of a virtual super-root is a good idea, because it introduces some
'node' to hold track information.

tpietzsch:
Cool. The key is "virtual", i.e., it should not be present in the graph and there should be no link from roots to the super-root, etc.

tpietzsch:
Regarding point (2) `Make (and use) a variant of the generic BreadthFirstIterator that is constructed with a Comparator for ordering children.`

I forgot that `LexicographicalVertexOrder` simply orders children (of the same root) by the edge sequence, so `BreadthFirstIterator` should already do the right thing. So I think that actually we should not do (2) for now to save the overhead of creating two `VertexKey`s for every comparison... IMO the best way to resolve this now is to order the list of roots and initialize the queue of the `BreadthFirstIterator` with the ordered sequence. The rest should just work (also for non-trees).

I just rewrote the `LexicographicalVertexOrder.VertexKey` construction from recursive to iterative to avoid StackOverflowErrors. So at least something came out of this detour... :-D

tinevez:
Yes I think that I mixed the `LexicographicalVertexOrder ` with `AlphaNumComparator`.

tinevez:
To address points 1 & 2, I added cross-component iterators to mastodon-graph.
See bigdataviewer/mastodon-graph#1

tinevez:
Hi @tpietzsch 
Can you check if this is OK?

Working on it I realized I could go one step further and make a search-action that would be graph-generic. Is it interesting for now?

PR #24, opened 2016-10-15
An importer for Trackmate files.

tinevez:
The importer can read the model as a whole and also import feature values.

The features keys are prepended with a prefix for spot and edge feature respectively to avoid name clashes.

An important impact of this PR is that TrackMate is no longer a dependency of the Mastodon. It can therefore have the license we want and we do not have to carry a list of dependencies as long as french income tax documentation.


tinevez:
Fixes #24

tpietzsch:
We should also have an exporter to be able to switch between MaMuT and Mastodon

PR #25, opened 2016-11-07
TrackScheme - Paint vertices and edges according to feature value.

tinevez:
TrackScheme can be painted according to feature values.


src/main/java/org/mastodon/revised/trackscheme/display/DefaultTrackSchemeOverlay.java
```diff
@@ -448,9 +704,98 @@ protected void drawVertexFull( final Graphics2D g2, final ScreenVertex vertex )
...
+		final TrackSchemeVertex tsv = graph.getVertexPool().getObject( vertex.getTrackSchemeVertexId(), ref );
+		final int[] incomingIDs = new int[ tsv.incomingEdges().size() ];
+		int i1 = 0;
+		for ( final TrackSchemeEdge tse : tsv.incomingEdges() )
+			incomingIDs[ i1++ ] = tse.getModelEdgeId();
+		final int[] outgoingIDs = new int[ tsv.outgoingEdges().size() ];
+		int i2 = 0;
+		/*
+		 * FIXME: Often, the JVM crashes on this line (or equivalent in the
+		 * other vertex methods) with the following error:
```
tinevez: It only occurs on windows, probably because the rendering is faster on this platform. I will investigate more.

tpietzsch: It also occurred for me on OS X. But on this branch, not on the model-edit-actions.

tpietzsch: Can you provide more details on your suspicion what could be the cause?

tinevez:
Superseded by the `screen-entitites-color` branch, that branched from this one and fixed a crash.

PR #35, opened 2017-09-05
Overlay graph renderer minor tweaks.

tinevez:
The content of this PR has been "manually" migrated to the `unify-views` branch by Tobias.

- [x] ` ScreenVertexMath` do not use AWT Ellipse class, but return ellipse parameters as a `double[]` array. We could even make it more general and remove the explicit dependency on `OverlayVertex` and just accept something that has a covariance and is `RealLocalizable`. I am sure this code is useful to a very wide audience.

- [x] Fix the `getVertexAt` method per @tpietzsch comments on how it should be in synch with what is painted. 

- [x] As @tpietzsch suggestion: moved the `ConvexPolytope` methods to utility class. 

- [x] Also made a PR in Imglib2 for symmetric matrix inversion system, so that we can remove this code from here.

- [x] Javadocs for methods and fields in OverlayGraphRenderer.

- [x] Fix and use the selection actions.  …
  - shift SPACE to select all edges/vertices in a track.
  - shift PAGE_UP for track upward selection.
  - shift PAGE_DOWN for track downward selection.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
 		<dependency>
 			<groupId>org.mastodon</groupId>
 			<artifactId>mastodon-collection</artifactId>
-			<version>1.0.0-beta-3-SNAPSHOT</version>
+			<version>1.0.0-beta-6-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>org.mastodon</groupId>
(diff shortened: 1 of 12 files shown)


tinevez:
Hi @tpietzsch 
So all the content of this PR has already been integrated in the main branch `unify-views`.
This PR and the corresponding branch can be safely deleted.

tpietzsch:
👍 thanks!

PR #36, opened 2017-09-12
User oriented feature support.

tinevez:
This PR adds feature value storage and facilities to compute them for the MaMuT model.

## Feature storage. 

 * General feature model stuff is in package `org.mastodon.revised.model.feature`.
 * The MaMuT `Model` has now a `FeatureModel`.

## Feature calculation.

 * For MaMuT, API consumers can extend special interfaces to implement classes able to provide these numerical features. These interfaces are discovered automatically by SciJava discovery mechanism. These are:
    * `org.mastodon.revised.mamut.feature.SpotFeatureComputer` and
    * `org.mastodon.revised.mamut.feature.LinkFeatureComputer`.
There are some examples in the same package.
 * A calculator class is in class or registering these computers, managing their dependencies, and feeding the `Model` with the results: `FeatureComputerService` with concrete impl. `DefaultMamutFeatureComputerService`.

## UI.

 * A GUI panel shows to the user what feature computers are available and let them select the feature they want: `org.mastodon.revised.mamut.FeatureComputersPanel`

## To discuss.

- [ ]  This PR imports SciJava and friends logic to the project. We probably want to make sure we do it in a better way than proposed now. 
 - [ ] It is probably a good time to also discuss feature serialization / deserialization, which is not touched by this PR.


src/main/java/org/mastodon/revised/model/feature/Feature.java
```diff
@@ -0,0 +1,129 @@
...
+		return projections;
+	}
+
+	/**
+	 * Returns the object this feature is defined on, as a
+	 * {@link FeatureTarget}.
+	 * 
+	 * @return the feature target.
+	 */
+	public FeatureTarget getTarget()
```
tpietzsch: Why is there a `FeatureTarget` enum when there is also a generic parameter `O` that defines the target?

tinevez: I could not come with one. 
After type erasure, the class does not expose anything that is `O`. So I deferred to explicit declaration with the enum.

tpietzsch: You could pass `Class<O>` as a constructor param instead of `FeatureTarget`?

src/test/java/org/mastodon/revised/model/mamut/io/TestSaveLoadFeatures.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.Iterator;
+
+import org.mastodon.pool.PoolCollectionWrapper;
+import org.mastodon.properties.DoublePropertyMap;
+import org.mastodon.revised.model.mamut.Model;
+import org.mastodon.revised.model.mamut.Spot;
+
+public class TestSaveLoadFeatures
```
tpietzsch: Is this incomplete or misnamed?

src/main/java/org/mastodon/revised/mamut/feature/DefaultMamutFeatureComputerService.java
```diff
@@ -0,0 +1,255 @@
...
+		{
+			logService.error( "Could not compute features using  " + computerNames +
+					" as they have a circular dependency." );
+			progressListener.showStatus( "Circular dependency!" );
+			return false;
+		}
+
+		final long start = System.currentTimeMillis();
+
+		model.getGraphFeatureModel().clear();
```
tpietzsch: This will cause problems with `manual` features? Complete `clear()` seems wrong here?

tinevez: True, noted.

src/test/java/org/mastodon/revised/model/mamut/io/TestSaveLoadFeatures.java
tinevez: To be removed. I could not get it right.

src/main/java/org/mastodon/revised/model/feature/Feature.java
tinevez: Noted! Thanks!

src/main/java/org/mastodon/revised/mamut/feature/LinkVelocityFeatureComputer.java
```diff
@@ -0,0 +1,68 @@
...
+	{
+		final ModelGraph graph = model.getGraph();
+		final DoublePropertyMap< Link > pm = new DoublePropertyMap<>( graph.edges(), Double.NaN );
+
+		final Spot ref1 = graph.vertexRef();
+		final Spot ref2 = graph.vertexRef();
+
+		@SuppressWarnings( "unchecked" )
+		final Feature< Link, Double, DoublePropertyMap< Link > > displacementFeature =
+				( Feature< Link, Double, DoublePropertyMap< Link > > ) model.getFeatureModel().getFeature( LinkDisplacementComputer.KEY );
```
tpietzsch: Would it be possible to make `FeatureModel.getFeature(...)` trigger computation if the feature is not present?
Might require to fuse `DefaultFeatureModel` and `DefaultFeatureComputerService`. This would provide automatic dependency resolution. 
(In any case, I would additionally keep to the explicit dependency resolution topological sort for the progress report and circular dependency detection.)

src/main/java/org/mastodon/revised/mamut/feature/DefaultMamutFeatureComputerService.java
```diff
@@ -0,0 +1,223 @@
...
+import org.scijava.app.StatusService;
+import org.scijava.log.LogService;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+import org.scijava.plugin.PluginInfo;
+import org.scijava.plugin.PluginService;
+import org.scijava.service.AbstractService;
+
+@Plugin( type = FeatureComputerService.class )
+public class DefaultMamutFeatureComputerService extends AbstractService implements FeatureComputerService< Model >
```
tpietzsch: There seems to be almost nothing MaMuT-specific about this. Pull out `Abstract(Default)ComputerService`?

src/main/java/org/mastodon/revised/model/feature/Feature.java
```diff
@@ -0,0 +1,128 @@
...
+ * @author Jean-Yves Tinevez
+ *
+ * @param <O>
+ *            the type of the object this feature is defined for (target).
+ * @param <T>
+ *            the type of the values of this feature.
+ * @param <K>
+ *            the type of the property map used to store feature values.
+ */
+public class Feature< O, T, K extends PropertyMap< O, T > >
```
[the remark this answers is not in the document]
tinevez: Adressed in dbef1e6d3faef644adad1ac870d8321339f2fc40

src/main/java/org/mastodon/revised/model/feature/FeatureComputerService.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Returns the collection of available computers for features defined on the
+	 * objects with the specified class.
+	 *
+	 * @param target
+	 *            the class of the objects the features are defined on.
+	 * @return the available feature computers.
+	 */
+	public Collection< FeatureComputer< Model > > getFeatureComputers();
```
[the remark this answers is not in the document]
tinevez: Addressed in 13527b1d013f7b0a91633267566ab961ccc633fb

src/main/java/org/mastodon/revised/model/feature/FeatureComputerService.java
```diff
@@ -0,0 +1,46 @@
...
+	 *
+	 * @param model
+	 *            the model to compute features on.
+	 * @param selectedComputers
+	 *            what computers to run.
+	 * @param progressListener
+	 *            a progress listener, used to report calculation progress.
+	 * @return <code>true</code> if computation terminated successfully.
+	 */
+	public boolean compute( AM model, Set< FeatureComputer< Model > > selectedComputers, ProgressListener progressListener );
```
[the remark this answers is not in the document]
tinevez: Addressed in 13527b1d013f7b0a91633267566ab961ccc633fb

src/main/java/org/mastodon/revised/mamut/feature/LinkVelocityFeatureComputer.java
tinevez: I need time to think about this.

src/main/java/org/mastodon/revised/mamut/feature/DefaultMamutFeatureComputerService.java
tinevez: Addressed in 1064e1003c8f02bc6447c1f3275958ee1cb60ffa

src/main/java/org/mastodon/revised/mamut/feature/LinkVelocityFeatureComputer.java
tinevez: I am not sure I feel great about going this direction.
I feel like the calculation and the accessing of features should be separated. Accessing a feature value should not trigger possibly crazy calculation, because this is the user responsibility to trigger calculation. We have the logic to inform the user that a feature value is not available for the object he is trying to access, and it his their responsibility to trigger calculation.

PR #37, opened 2017-11-12
Feature de/serialization. DO NOT MERGE

tinevez:
- Feature de/serialization is triggered from the `FeatureModel`, when the
model is being serialized itself.

- Features are de/serialized from/to a subfolder of the project folder.
With the `DefaultFeatureModel,` this subfolder is 'Features'.

- Each feature is saved to an individual file, named with the key of the
feature + '.raw'.

- Specialized classes, derived from the interface `FeatureSerializer`,
are in charge of the de/serialization of individual features. There is
*one* specific `FeatureSerializer` for each feature. This is necessary
since we do not know what is the type of values used by a feature.

- Concrete `FeatureSerializer` implementations are provided by the
`FeatureComputer`s. Since these classes (normally written by 3rd party
users) are the ones that know how to generate a feature, they also know
how to de/serialize them.

- So the model needs to know what are the `FeatureComputer`s when saving
and loading. Since they are provided by SciJava plugin discovery mechanism
(within the FeautureComputerService), the model needs to access a SciJava
`Context`. Now each `Model` instance has its own context (lightweight,
only instantiated with the `MamutFeatureComputerService`). The context
instance has then been removed from the `MainWindow` class, and the later
uses the context of the model it manages.

### DIFF
--- a/src/main/java/org/mastodon/revised/mamut/FeatureComputersPanel.java
+++ b/src/main/java/org/mastodon/revised/mamut/FeatureComputersPanel.java
@@ -66,7 +66,7 @@ public class FeatureComputersPanel extends JPanel
 
 	private final MyProgressBar progressBar;
 
-	private final Set< FeatureComputer< Model > > selectedComputers;
+	private final Set< FeatureComputer< ?, ?, Model > > selectedComputers;
 
 	private final JButton btnCompute;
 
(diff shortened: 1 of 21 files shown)


tinevez:
Let's redo this after the changes in the `Feature` framework are merged.

PR #38, opened 2017-11-15
TrackScheme zoom behaviour.

tinevez:
Drag a rectangle over TS with the 'Z' key. TS zooms to this rectangle.
A press of 'Z' fully zoom to deisgnated location.
If TS is fully zoomed, a tap of 'Z' unzoom fully.

trackschemezoombehavior


tinevez:
Let's wait for the plugin framework.

tinevez:
Would you prefer I do a plugin from this? 
I still need to change/add a couple of methods in the core, but the zoom itself could be a `Mastodon plugin`.

PR #39, opened 2017-12-04
Add a config page in Mastodon to manage a collection of BDV render settings.

tinevez:
Like this:
capture

This strongly emulates the logic found for `TrackSchemeStyles`.

Careful: Changes to and new RenderSettings created in the preference dialog are not used nor visible in BDV windows yet.


tinevez:
Trashed?

PR #47, opened 2018-02-14
Add an action to close a view window.

tinevez:
To properly close the view, we send it a` WINDOW_CLOSING` event.
This way, the listeners of the` JFrame` are called and the closing happens gracefully within Mastodon.

### DIFF
--- a/src/main/java/org/mastodon/app/ui/CloseWindowActions.java
+++ b/src/main/java/org/mastodon/app/ui/CloseWindowActions.java
@@ -1,0 +1,77 @@
+package org.mastodon.app.ui;
+
+import static org.mastodon.app.ui.MastodonFrameViewActions.CLOSE_WINDOW;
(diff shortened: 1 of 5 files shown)


tpietzsch:
You set keystroke `ctrl W`. It should be `meta W` (==`cmd W`) for MacOs. And I would add the respective default keystrokes for Windows and Linux (Ubuntu or whatever standard flavour...). 

...maybe check what IJ does on other platforms. It has cmd-W for closing Images on MacOs...

tinevez:
Yes.

tinevez:
Can we add 'per-platform' keystroke?

tpietzsch:
Sounds difficult to fit into the ui-behaviour. The way I would do it is that we provide default keymaps for different platforms. And then pre-select the appropriate default for the platform the first time mastodon is started.

tinevez:
It's Ctrl-W on Windows platforms and Linux.

tpietzsch:
I added close actions for `PreferencesDialog` and `TagSetDialog`.

PR #49, opened 2018-02-22
Search vertices by their label.

tinevez:
This PR adds a `SearchVertexLabel` action that is added to the BDV and TrackScheme views (last item on their settings panel). It shows up as a search field with a checkbox for options. `Ctrl F`, `meta F`, `/` move the focus to the search field. 

Search starts from the selected vertex if there is exactly one. If not, it starts from the focused vertex. If no vertices are focused, it starts from the first one.

Search processes in breadth-first order, connected-component by cc. Iterated CCs are sorted by their label, as in TrackScheme.  

This PR requires several changes:
- `TrackSchemeVertex` now implements `HasLabel` and `HasTimepoint`. It does in practice, but we need the interface to be declared.
- The `ViewFrame`s expose their settings panel via the `getSettingsPanel()` method. This is needed to added elements to this settings panel such as the search field.
- This PR is temporary coupled to mastodon-graph 1.0.0-beta-11-SNAPSHOT because I made a mistake and the graph iterators accepted a `Graph` instead of a `ReadOnlyGraph`.


tinevez:
Bugs to fix and improvements, noted by @tpietzsch:
- [ ] `ESC` should release the focus from the search field.
- [ ] Problem with the search wrapping at the end: it gets stuck to the last found vertex and does not loop. It should loop when it has exhausted all vertices, and indicate the wrap-around by a special icon.
- [ ] The search should start from the last focused vertex.

PR #50, opened 2018-02-23
Locking within overlay-graph-renderer.

tinevez:
Methods that call `getVertexAt` and `getEdgeAt` are responsible for locking the graph. 
This PR checks that it is the case.


tpietzsch:
👍

PR #51, opened 2018-02-23
 Ask the user to save the project before closing the main window.

tinevez:
capture

### DIFF
--- a/src/main/java/org/mastodon/revised/mamut/MainWindow.java
+++ b/src/main/java/org/mastodon/revised/mamut/MainWindow.java
@@ -27,6 +27,8 @@ import org.mastodon.revised.ui.keymap.Keymap;
 
 public class MainWindow extends JFrame
 {
+	private static final long serialVersionUID = 1L;
+
 	protected final JMenuBar menubar;
 
 	private final ViewMenu menu;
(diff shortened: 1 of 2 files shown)


tpietzsch:
This needs more work:

The "Save before close?" dialog should only appear if
1.) there is a project open, and
2.) it has been modified since it was last saved.

If the user then clicks "Save project", and after that cancels the Save project dialog, then Mastodon should not close (this should be the same as if the user had clicked "Cancel" in the "Save before close?" dialog.

tpietzsch:
I'll add functionality to the `UndoRecorder` to set "save points" and query wether you are currently at one. Then there will be no need to listen for graph changes. (which anyway gives only an incomplete picture, because it doesn't capture for example changing the tag set structure).

Follow-up request: Add a similar dialog when loading a project or creating a new one: "Current project is not saved. Do you want to..." bla bla.

tinevez:
> Follow-up request: Add a similar dialog when loading a project or creating a new one: "Current project is not saved. Do you want to..." bla bla.

See my comment on #53

tinevez:
Superseded by https://github.com/mastodon-sc/mastodon/pull/168

PR #52, opened 2018-02-27
Feature-based coloring for TrackScheme views.

tinevez:
capture


tpietzsch:
I have not reviewed yet, but I have rebased it on current master. The result is in branch `feature-coloring-rebased`.

tpietzsch:
- [ ] I don't like that the "Feature Calculation" panel is in the Preferences dialog. It is project specific.
For now, I would simply put it in its own dialog. Use [TagSetDialog](https://github.com/fiji/TrackMate3/blob/master/src/main/java/org/mastodon/revised/model/tag/ui/TagSetDialog.java) as a template. That is, keep it the panel as a SettingsPage and put it into a [SingleSettingsPanel](https://github.com/fiji/TrackMate3/blob/1cb7471a9e932ec9e94440258be7e092be9551b8/src/main/java/org/mastodon/revised/model/tag/ui/TagSetDialog.java#L109). Trigger it from the (currently unused) "features and tags" button in the MainWindow
screen shot 2018-03-02 at 22 16 56

- [ ] I'm not sure about this, but I think it would be better to not use a `SelectAndEditProfileSettingsPage` for the "Feature Color Modes". The inconsistency is, that there is no "select" here. It is wired up somehow, but it doesn't make much sense. The Coloring menu is used to "select", and it can be different for each window.
Consider using something like the `TagTable` in the tag sets dialog
screen shot 2018-03-02 at 22 22 42I imagine a table of the coloring modes on the left and the mode configuration panel (for the selected entry) on the right. Then this whole thing is put into the Preferences dialog as a page.
`TagTable` is maybe a bit difficult to understand. Basically, the [constructor](https://github.com/fiji/TrackMate3/blob/master/src/main/java/org/mastodon/revised/model/tag/ui/TagTable.java#L10-L17) wires it up to a model with lambdas that define an "interface" on-the-fly.
Look at [TableDialogExample](https://github.com/fiji/TrackMate3/blob/master/src/test/java/org/mastodon/revised/model/tag/ui/TagTableDialogExample.java). Selection and addition of elements can be tracked by listeners, see [here](https://github.com/fiji/TrackMate3/blob/master/src/main/java/org/mastodon/revised/model/tag/ui/TagSetPanel.java#L62) and [here](https://github.com/fiji/TrackMate3/blob/master/src/main/java/org/mastodon/revised/model/tag/ui/TagSetPanel.java#L64)
Later I would rename `TagTable` because it is not exclusively for tags anymore. But for now, just use it. Also, later we might want to extend `TagTable` with a "duplicate" button on each line, which duplicates that entry into a new one. For now, just let users define new coloring mode from scratch.

tinevez:
Superseded and rebased on the `master-feature-coloring` branch.

PR #55, opened 2018-06-03
Re-engineer feature framework.

tinevez:
This PR implements a series of API-breaking changes to sanitize and improve the feature framework. 
They will affect the `mastodon-tracking` code, because particle detectors and linkers generate features themselves. 

Changes in the common `Feature` API.
------------------------------------------

The main change is that now **`Feature`** is an **interface**, that ships an API making it resemble a **read-only version of a property map**, plus commodities to play well with end-users, whether they use features to harvest quantitative data on the image analyzed, or extend Mastodon by writing Feature computers.

A `Feature` can directly return the values it wraps, using 
``` java
public V get( O o, V ref );
```
where `V` is the type of values and `O` the type of objects the feature is defined for (vertices, edges, ...).

A feature should also be able to **serialize** itself to a file, following what we do with property maps. 
``` java
public void serialize( File file, final ObjectToFileIdMap< O > idmap ) throws IOException;
```

Because values can be of any type and we want to let 3rd parties compute their own features, this is not super-useful for integration within Mastodon. We work around this by providing **`FeatureProjection`s.** Feature projections are meant to look like the scalar projections of a feature value on several axis. For instance, the velocity vector is a feature calculated for an edge. Adequate feature projections for this feature would be: 
- the speed (velocity vector magnitude). 
- the signed velocity along X axis. 
- the signed velocity along Y axis. 
- the signed velocity along Z axis. 
- theta the velocity vector angle measured in the OXY plane with respect to X axis (0º to 360º). 
- phi the velocity vector angle measured in the OMZ plane with respect to Z axis (0º to 180º). 

The last 6 quantities are scalar and real. Note that they are not orthogonal; end-user comfort should prime. Because they are always scalar and real, they can be used in a general way, _e.g._ to generate colors for displays or in a table.

Feature projections have **physical units** now. There is also a `IntFeatureProjection` marker interface that can be used for feature projections that return integer values, and will be used for instance to beautify the output of values.

There concrete classes to simplify generating features, covering common uses cases:
- `DoubleScalarFeature`
- `DoubleArrayFeature`
- `IntScalarFeature`
- `IntArrayFeature`

Changes in the common feature computation.
---------------------------------------------------

Since a `FeatureComputer` is like a self-contained module that can be contributed by 3rd parties, we should try to make them user-friendly. 

Now they need to return a **help string** that explains what the feature is about and how it is calculated. For TrackMate I did not do that and still receive a lot of emails and posts in the forum about what is the exact definition of a quantity. Of course, feature values are what users make sense of their data with and they want to know what is behind a value before putting it in the paper. 

Also, a `FeatureComputer` can return a **config panel**. This makes it no longer state-less, but will be useful for special features.

Changes in the Mastodon-app (Mamut) feature computation.
--------------------------------------------------------------------

We want to be able to write features that depend on image data, for instance the mean intensity in spots. To do this, the Mamut feature computers need to receive the BDV data. The **`MamutFeatureComputer`** does this and should be the mother class for all Mastodon-app feature computers. It also determines the space units from the BDV data.

The `DefaultMamutFeatureComputerService` passes the BDV data to all feature computers before computation. 

Mastodon-app feature calculation dialog.
----------------------------------------------

The calculation dialog exploits all these capabilities. Here is how it looks like, with a feature computers that can be configured. Clocking on the gear icon next to a feature computer shows an info and config panel on the right pane.
mastodon_featurecalculation_01

It also shows dependencies when any:
mastodon_featurecalculation_02

The dialog is now present in the Mastodon app `MainWindow`
mastodon_mainwindow

and there are a new command to toggle its visibility:
``` java
public static final String COMPUTE_FEATURE_DIALOG = "compute features";
```


tpietzsch:
I would like to rename `IntScalarFeature` to `IntFeature` and `DoubleScalarFeature` to `DoubleFeature` to be more consistent with attributes (where there is `IntAttribute`, `IntArrayAttribute`, ...)

tpietzsch:
I don't like that serialization is split between `Feature` and `FeatureComputer` (with `Feature` doing serialization and `FeatureComputer` doing deserialization).
I would remove serialization for now and make a separate PR when it is improved.

tpietzsch:
I removed serialization in https://github.com/fiji/TrackMate3/pull/62 (PR to this PR...)

tpietzsch:
`AbstractFeatureComputerService.getDependencyGraph()` is broken.
`prune()` will not work as expected. Here is an example:
[Mastodon (1).pdf](https://github.com/fiji/TrackMate3/files/2311430/Mastodon.1.pdf)

tinevez:
Regarding pruning of missing dependencies: 
The user can change what features they want computed at every run. For instance they might want to compute features A and B with B depending on A.
If in a subsequent call they remove A to just compute B, we need to re-analyse dependencies to put back A.

tinevez:
Regarding `AbstractFeatureComputerService.getDependencyGraph()`:
I will work on it.

tpietzsch:
Sure, but we could prune the full dependency graph before.
Which dependencies are missing and therefore which features can never be computed (whatever the user selects) can be determined at initialization.
With the pruned full dependency graph, the per-`compute()`-necessary `getDependencyGraph()` no longer has to prune anything, because everything is well-defined.

tpietzsch:
Regarding dependency graph, I would:

1) Add a boolean `valid` flag to nodes. You could make your own `AbstractObjectVertex` subclass, similar to `ObjectVertex`
2) build a full dependency graph of all registered `FeatureComputer`s
if you encounter a missing dependency, add a node for it anyway, and set it's valid flag to false
3) after the graph is build, propagate valid==false backwards to all parents
4) remove all nodes that have valid==false from the graph
These `FeatureComputer`s will never run. They can also be removed from the `featureComputers` map.

Same thing for loops: If there is a cyclic dependency loop, no FeatureComputer on that loop will ever run.

Then in the `getDependencyGraph()` you can basically do what you do now, but it will not happen that FeatureComputers are missing. And it will also never happen that there is a cyclic dependency.

PR #56, opened 2018-06-05
Feature-based coloring for TrackScheme, with user-configurable modes.

tinevez:
Please review and merge #55 first.

A settings page to edit color modes.
------------------------------------

User-provided information about how to generate colors based on feature values is stored in a `FeatureColorMode` class, that can be serialized to YAML, and is managed like the BDV `RenderSettings` and the `TrackSchemeStyle`.

The mode editing panel resembles this:
mastodon_featurecolormodepanel_01

Classic case like coloring a vertex by a vertex feature is supported, but it is also possible to color and edge by a vertex feature and vice-versa:
mastodon_featurecolormodepanel_02

The editor looks like this on Windows platform:
mastodon_featurecolormodes_01

The color maps are represented by a `ColorMap` class, that can also be serialized to YAML. There is no editor for these colormaps, however it is possible to drop ImageJ LUT files in the `~/.mastodon/luts` folder:
``` sh
tinevez@lilium:~$ ls .mastodon/luts/
Algae.lut   Curl.lut    Dense.lut   Ice.lut     Phase.lut   Tempo.lut
Amp.lut     Deep.lut    Gray.lut    Matter.lut  Solar.lut   Thermal.lut
Balance.lut Delta.lut   Haline.lut  Oxy.lut     Speed.lut   Turbid.lut
```

``` sh
tinevez@lilium:~$ cat .mastodon/luts/Algae.lut 
Index	Red	Green	Blue
0	17	36	20
1	17	37	20
2	18	37	21
3	18	38	22
4	18	39	22
...
```

The coloring is based on the **feature projections** (scalar, double values) of a feature. The editor panel lets the user pick a pair of feature and projection keys:
mastodon_featurecolormodepanel_03

Finally, the validity of a color mode depends on the feature to be calculated. It is possible and even common to have a color mode pointing to features that are not registered in the feature model. We want to let the user edits such modes, but we warn them that the related features need to be computed before use in a view:
mastodon_featurecolormodepanel_04

TrackScheme coloring.
-----------------------

The TrackScheme coloring menu now includes feature-based color modes. The color modes that are invalid (depending on feature that are not yet computed) appear disabled in the menu:
mastodon_tracschemecoloring_01

It looks like this on Windows platform:
mastodon_trackscheme_featurebasedcoloring_01

Here is an example:
mastodon_tracschemecoloring_02

Another example for a track feature: all the spots of a track have the same feature values, ergo the same color:
mastodon_tracschemecoloring_03


src/main/java/org/mastodon/revised/model/feature/Feature.java
```diff
@@ -33,94 +36,77 @@
  *
  * @param <O>
  *            the type of the object this feature is defined for (target).
- * @param <K>
- *            the type of the property map used to store feature values.
+ * @param the
```
imagejan: This throws a javadoc error now. Should be:

    * @param <V>

tinevez:
Superseded by PR now merged.

PR #57, opened 2018-06-05
A view of the data as a table.

tinevez:
Please review and merge #56 before merging this PR.

View model as a simple table.
-----------------------------

A new view is added by this PR. It reuses the `MamutView` framework already used for the BDV and TrackScheme views.  

There is two flavors of the table view, that can be found in the `Window` menu:
mastodon_table_03

The data table shows the full model as a two tables. They list object label, id, feature values and tags. There is one table for vertices:
mastodon_table_01
And one for edges:
mastodon_table_02

The columns are updated live when features are computed or when the tag structure is changed. The rows are changed when the graph is edited. 

Selection, Focus and Highlight view.
------------------------------------

The table is in sync with the selection model:
mastodon_table_04

And with the focus and the highlight models. The focused object is shown with a special font. The highlighted object with a thick border:
mastodon_table_05

Table navigation.
-----------------

Using the group locks, the table can receive and send navigation events:
mastodon_table-navigation

Table editors.
--------------

The table lets the user edit tags and the vertex labels:
mastodon_table_06

mastodon_table-edit

Table context.
--------------

The table is a context listener. When a BDV context is chosen, it only displays the context content at current time-point:
mastodon_table-context

Selection Table.
----------------

The second flavor of the table is the selection table. It only displays the current selection and does not edit it:
mastodon_table-selection-table

API changes.
-------------

For the context handling, I added a method in the `Context` interface that returns the current time-point:
```java
public interface Context< V >
{
	public Lock readLock();

	public Iterable< V > getInsideVertices( final int timepoint );

	public int getTimepoint();
}
```

Tags and feature-based coloring.
------------------------------------

There is now tag-set and feature based coloring for the table views:

bl8w8-6n64z


tinevez:
Superseded by #68

PR #61, opened 2018-08-22
Zip project

tpietzsch:
Store the project in a single file instead of directory.
This makes moving projects around easier, as well as selecting them in file dialogs.
Project files have `.mastodon` extension but are simply (uncompressed) zip files.
(Later we could open Mastodon automatically when a `.mastodon` is drag-and-dropped into Fiji.


src/main/java/org/mastodon/revised/model/mamut/trackmate/MamutExporter.java
```diff
@@ -681,7 +681,9 @@ public static void main( final String[] args ) throws IOException
 		final String bdvFile = "samples/datasethdf5.xml";
 		final MamutProject project = new MamutProject( new File( projectFolder ), new File( bdvFile ) );
 		final Model model = new Model();
-		model.loadRaw( project );
+		final MamutProject.ProjectReader reader = project.openForReading();
+		model.loadRaw( reader );
+		reader.close();
```
tinevez: Could we make this compatible with the `try-with-resources` construct?
We pass input and output streams around but they are still `AutoCloseable` so maybe something like
```java
try(final MamutProject.ProjectReader reader = project.openForReading())
{
   model.loadRaw( reader );
}
```
would normally work.

tpietzsch: Sounds good!

tinevez: Let's go!

PR #63, opened 2018-10-12
Updates to Tobias new feature framework.

tinevez:


src/main/java/org/mastodon/mamut/feature/SpotGaussFilteredIntensityFeatureComputer.java
```diff
@@ -166,11 +214,54 @@ public void run()
...
+		private final Map< Integer, Collection< Spot > > index;
+
+		public MyIndex( final GraphUpdate< Spot, Link > update )
+		{
+			this.index = new HashMap<>();
+			for ( final Spot spot : update.vertices( UpdateLocality.SELF ) )
+			{
+				final int timepoint = spot.getTimepoint();
+				Collection< Spot > iterable = index.get( Integer.valueOf( timepoint ) );
+				if ( null == iterable )
```
tpietzsch: Can be replaced with `computeIfAbsent`.
Integer.valueOf() is not needed.

src/main/java/org/mastodon/feature/ui/FeatureComputationController.java
```diff
@@ -0,0 +1,141 @@
...
+		gui.btnCompute.addActionListener( ( e ) -> compute() );
+		gui.btnCancel.addActionListener( ( e ) -> cancel() );
+
+		gui.progressBar.setString( "" );
+		propertyChangeListener = new PropertyChangeListener()
+		{
+			@Override
+			public void propertyChange( final PropertyChangeEvent evt )
+			{
+				if ( "progress".equals( evt.getPropertyName() ) )
```
tpietzsch: It seems that "progress" event is never sent anywhere.
What is the idea behind this?

src/main/java/org/mastodon/mamut/feature/SpotGaussFilteredIntensityFeatureComputer.java
tinevez: Noted.

src/main/java/org/mastodon/feature/ui/FeatureComputationController.java
tinevez: The `PropertyChangeListener` is an object that is passed to _feature computers_  as a dependency.
They are in charge of reporting `progress` if they want. An example is the gaussian mean intensity computer. Because it takes much much longer that all the other computers, it makes sense for it to report progress.

src/main/java/org/mastodon/mamut/feature/SpotGaussFilteredIntensityFeatureComputer.java
```diff
@@ -166,11 +214,54 @@ public void run()
...
+				{
+					iterable = new ArrayList<>();
```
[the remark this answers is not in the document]
tinevez: Correct!

PR #71, opened 2018-11-28
Simplify and generalize focus model.

tinevez:
- Remove unused generic type (edges)
- Generalize: the focused object does not have to be a vertex in a graph.


tinevez:
Defer this to another attempt.

PR #74, opened 2018-12-06
Set enabled flag of selection menu items from EDT thread.

tinevez:
Ok, so this is downstream a bug that caused Mastodon to freeze completely when using the semi-automatic tracker.

When the semi-automatic tracker is launched from a 'normal' thread, it completely froze Mastodon. The reason is that the tracker changes the selection model when a spot is created or discovered. The selection model listeners in the `SelectionActions` class that toggle whether some menu items (related
to selection) are enabled or not.
Since the listeners were not triggered from the EDT, this caused the freeze.

To fix this problem, this commit simply ensures that the `setEnabled` method from related `Action` is called from the EDT with `SwingUtilities.involkeLater()`.

### DIFF
--- a/src/main/java/org/mastodon/revised/ui/SelectionActions.java
+++ b/src/main/java/org/mastodon/revised/ui/SelectionActions.java
@@ -3,6 +3,8 @@ package org.mastodon.revised.ui;
 import java.awt.event.ActionEvent;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
+import javax.swing.SwingUtilities;
+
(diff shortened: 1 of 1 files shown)


tpietzsch:
Thanks for noticing this!!!!

Maybe later we can investigate:
 * whether `SwingUtilities.invokeLater()` needs to be used on construction of the action (probably not)
 * whether it is possible to only use `invokeLater()` if the enabled state changes (and do nothing otherwise), and whether that has any performance benefit.

Issue #66, opened 2018-10-17
missing 0th frame in .xml after "Export mamut"


xulman:
Hi,
[this dataset](http://www.fi.muni.cz/%7Exulman/files/secret_folder/missing0thFrame.zip) consists of frames t1.tif till t9.tif with timepoints 1 till 9 (according to its DBV's `dataset.xml`. However, when opened as "New project" in Mastodon,  the timepoints are treated 0 till 8 (in BDV as well as in TS). Now, when exporting into Mamut's .xml, the spot definitions are with timepoints 1 till 9, e.g.
```<SpotsInFrame frame="9" />```,
and definitions of spots from the first frame are missing (but their ID's are referenced from definitions of tracks). Also the
```<AllSpots nspots="154">```
counts all spots and is larger than number of lines with individual spot definitions in the created Mamut's .xml.

**Comment by xulman** [comment]
It gets even more funky when input dataset.xml consists of timepoints e.g. 5,6,7 (then 0,1,2) is (attempted to get) exported...

**Comment by xulman** [comment]
I've patched that in a very nasty way with [these 3 commits](https://github.com/xulman/TrackMate3/commits/patchingMamutExport), check out their commit messages

**Comment by tinevez** [comment]
Hi @xulman 
Tobias and I rewrote a very large chunks of mastodon core and merged it to master recently. 
Since this work focused on numerical features, there are little chance that it fixed this bug, but just to be sure, would you mind trying again?
Thanks!

**Comment by tinevez** [comment]
Yes.

**Comment by tinevez** [comment]
@xulman I am lazy, can you re-upload the dataset that allows to test the solution? Thanks.

**Comment by xulman** [comment]
Hi @tinevez,
thanks for working on it... [here's](http://www.fi.muni.cz/~xulman/files/secret_folder/testingMastodon.zip) the new testing folder:

Dataset is defined on timepoints 5 till 8, and the Mamut export seems to keep trying (polling) some other timepoints as the output .xml contains this:
```
    <AllSpots nspots="8">
      <SpotsInFrame frame="5" />
      <SpotsInFrame frame="6" />
      <SpotsInFrame frame="7" />
      <SpotsInFrame frame="8" />
    </AllSpots>
```

xulman:
BTW: have you tried in Mastodon to add linked spot beyond the last frame?
[...]

PR #77, opened 2018-12-12
FIX: Mamut XML exporter:

xulman:
deals with issue #66 

When exporting into Mamut format:
- time points are fetched from the underlying data (e.g. dataset.xml),
  and they order themselves in a, not necessarily continuous, sequence
  that may start, not necessarily, from non-zero time value
- Mastodon internal data, however, store time points zero-based sequentially
- both series must be of the same length
- the fix just iterates them in synchrony


tpietzsch:
@xulman There is already a branch `fix-mamut-frames` by JY, which has this fix + one additional fix for a different problem. I'll merge that to master now. Sorry for not being faster and the additional work ...

tpietzsch:
Ok, done, it's on master. I'll make a new release including the fix next week.

xulman:
no problem! thank you for fixing it... Mastodon flourishes ;)

PR #78, opened 2018-12-13
ADD: ability to adjust graph model, not only reset it.

xulman:
The standard way to pause model listeners and "lock" the model for changes is via
```
new AbstractModelImporter< Model >(pluginAppModel.getAppModel().getModel()) {{ startImport(); }};
//my code here
new AbstractModelImporter< Model >(pluginAppModel.getAppModel().getModel()) {{ finishImport(); }};
```
But `startImport` involves cleaning of the model... since I was blind to find how to only *update* the model without cleaning it first, I propose this PR.

### DIFF
--- a/src/main/java/org/mastodon/mamut/importer/ModelImporter.java
+++ b/src/main/java/org/mastodon/mamut/importer/ModelImporter.java
@@ -13,6 +13,30 @@ public class ModelImporter extends AbstractModelImporter< Model >
 		this.model = model;
 	}
 
+	/**
+	 * Pauses listeners attached to graph and tags. Graph and tags can be updated silently.
+	 * Unpause with {@link #finishUpdate()}.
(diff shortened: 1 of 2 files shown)


xulman:
Hi, I've revamped this piece.
It's been rebased to the latest master.

I could use such "official" way to pause listeners when *updating* graph, the original code was always clearing the graph.
V.

tinevez:
As suggested by Tobias elsewhere, another option would be to allow for controlling the listeners. Like have something like:

```java
model.listeners().pause(); // pseudo code hein
try
{
    // Add, remove, do plenty of things in bulk
}
finally
{
    model.listeners().resume(); // fire listeners if a change happen.
}

```

What would you favor?

tinevez:
@xulman do you still need this?

PR #79, opened 2018-12-16
An OverlayRenderer that displays a color-bar for the FeatureColorModel on Mastodon views.

tinevez:
The color-bar looks like this on TrackScheme, when a tag-set is selected in the coloring menu:
mastodon_colorbar_05

Zooming a bit more:
mastodon_colorbar_04

For feature-based coloring, when the parameters are identical for vertices and links:
mastodon_colorbar_03

And when they are different:
mastodon_colorbar_02

A sub-menu of the view menu, just under the coloring menu, controls the visibility and position of the color-bar:
mastodon_colorbar_01


tinevez:
Integrated since beta-21 and superseded for bete-22

PR #80, opened 2018-12-17
 Fix trying to add and link beyond last time-point.

### LINKED ISSUE #75: TS adds new timepoint beyond the dataset timepoints span
- while annotating a dataset, move to the last timepoint,
- move mouse over some spot in BDV,
- press A (add linked spot) and it gets a bit crazy... would expect nothing to happen actually

TS seems to add another time point (a new line appears in TS) but does allow to see it (to scroll over it). BDV shows ellipses from the last time point and also from the new just-beyond time point. At least, this is what it looks like from outside.

How can I help?

tinevez:
Fix trying to add and link beyond last time-point. 
Or before first time-point.
Reported by @xulman
Fixes #75

Also deal with moving the time slider while adding spots.


tinevez:
Fixed.

PR #87, opened 2019-03-08
ADD: ProjectManager.getProjectTitle() reports project path

xulman:
Hi,
was just missing a piece of API to see the current project's title...
V.


xulman:
[here's how to use example: into plain txt exporter...](https://github.com/xulman/CTC-FijiPlugins/blob/147ac21dcf7f9e2b50c5cec799d0701c524d35af/CTC-related/src/main/java/de/mpicbg/ulman/Mastodon/PlainTextFileExport.java#L319)

tpietzsch:
I added ProjectManager.getProject() instead

PR #89, opened 2019-03-11
Edit Tags in the menu and with a key listener.

tinevez:
The edit menu now has a submenu named `tags` that allow for settings the tags of the current selection. 

Also, the `EditTagActions` can now harness a number of tagsets and tags larger than 9. Just compose the number you want to select by pressing the key in quick succession. 

If there are less than 9 tags or tagesets, the previous, classical, immediate key listener is used.

### DIFF
--- a/src/main/java/org/mastodon/revised/util/GroupStrings.java
+++ b/src/main/java/org/mastodon/revised/util/GroupStrings.java
@@ -15,24 +15,24 @@ import java.util.Set;
  * words.
  * <p>
  * For instance, the 3 strings:
- * 
+ *
  * <pre>
- * - "Save Project"
- * - "Import TGMM tracks"
- * - "Import Simi BioCell tracks"
+ * -"Save Project"
+ * -"Import TGMM tracks"
+ * -"Import Simi BioCell tracks"
  * </pre>
  *
  * will be grouped in 2 groups:
- * 
+ *
  * <pre>
- * - "Save Project"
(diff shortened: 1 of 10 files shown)


src/main/java/org/mastodon/revised/util/GroupStrings.java
```diff
@@ -0,0 +1,252 @@
...
+			split.addAll( sg.split() );
+		
+		return split;
+	}
+
+	private static List< Group > fuseCollection( final Collection< Group > c )
+	{
+		final Deque< Group > champions = new ArrayDeque<>( c );
+		final List< Group > fused = new ArrayList<>();
+		if ( champions.size() > 1 )
```
tpietzsch: If input contains 1 group, `fuseCollection` will return an empty list. Is this a bug?

src/main/java/org/mastodon/revised/util/MnemonicsAssigner.java
```diff
@@ -0,0 +1,70 @@
...
+	{
+		this.buttons = new HashMap< >();
+	}
+
+	public void add( final AbstractButton button )
+	{
+		buttons.put( button.getText(), button );
+	}
+
+	public void assignMnenonics()
```
tpietzsch: typo: assignMne__n__onics

tpietzsch:
nice!

PR #101, opened 2019-04-04
 Animate time in BDV windows with navigation events.

tinevez:


tinevez:
Our attempts did not result in something good.
The animation becomes super berserk if there are several views linked together.
We probably miss a more complex animation behavior.

Issue #96, opened 2019-03-18
Special focus visibility mode for BdvOverlay.


tinevez:
A mode for BdvOverlay where ellipsoids are only shown for the track that contains the focussed vertex

PR #102, opened 2019-04-12
BDV visibility modes.

tinevez:
To use them, press `V` in a BDV window.

4 modes are implemented:
- ALL: everything is visible.
- NONE: nothing is visible.
- FOCUS: the track of the focused vertex is visible.
- SELECTION: the current selection is visible.

The Visibility objects are used every-time one has to decide on the visibility of a vertex or an edge. This includes calls to methods like `#getVertexAt()` or `#getEdgeAt()`. Also the methods called by
the `BdvContextProvider`, to keep visibility consistent.

The visibility itself is not super clever. We simply rely on the `isVisible()` method inside for loops over the vertices and edges returned by the clip convex polytope.


tinevez:
Fixes #91 
Fixes #96

tinevez:
Integrated since beta-21

PR #111, opened 2019-12-13
Let the user select what sources they want to compute feature on.

tinevez:
The features that have a `ON_SOURCES` mutiplicity can now be configured to compute the feature only on certain sources. 

It looks like this in the feature computation dialog:
Screen Shot 2019-12-13 at 13 44 36

There is a model class that stores what the user selects, and that can be added as dependency in feature computation:
Screen Shot 2019-12-13 at 13 46 22

From this class, the computer must retrieve the specific settings for the feature they compute:
Screen Shot 2019-12-13 at 13 47 42

And use it in the computation loop:

Screen Shot 2019-12-13 at 13 48 30

### DIFF
--- a/src/main/java/org/mastodon/feature/FeatureComputationSettings.java
+++ b/src/main/java/org/mastodon/feature/FeatureComputationSettings.java
@@ -1,0 +1,117 @@
+package org.mastodon.feature;
+
+import java.util.BitSet;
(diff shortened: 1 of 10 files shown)


tinevez:
Give it up. 
Let's not configure feature computers, but instead make custom computers.

PR #120, opened 2020-05-23
ADD: MamutViewBdv.getViewerPanelMamut() getter to be able to listen

xulman:
for, e.g. view transform change or time point change


xulman:
Hi,
thank you!
Vlado

PR #129, opened 2021-04-10
Add BdvViewCreatedListener

ksugar:
This PR adds the following functionalities to `org.mastodon.mamut.WindowManager`.

- Classes that implement `BdvViewCreatedListener` get a notification when a new `MamutViewBdv` instance is created.
- `getBdvWindows()` exposes currently open BigDataViewer windows.

### DIFF
--- a/src/main/java/org/mastodon/mamut/WindowManager.java
+++ b/src/main/java/org/mastodon/mamut/WindowManager.java
@@ -187,6 +187,8 @@ public class WindowManager
 
 	final ProjectManager projectManager;
 
+	private final List< BdvViewCreatedListener > listeners;
+
 	public WindowManager( final Context context )
 	{
 		this.context = context;
@@ -246,6 +248,8 @@ public class WindowManager
 		globalAppActions.namedAction( tooglePreferencesDialogAction, PREFERENCES_DIALOG_KEYS );
 
 		updateEnabledActions();
+
+		listeners = new ArrayList<>();
 	}
 
 	private void discoverPlugins()
@@ -385,6 +389,7 @@ public class WindowManager
 			final MamutViewBdv view = new MamutViewBdv( appModel, guiState );
 			view.getFrame().setIconImage( BDV_VIEW_ICON );
 			addBdvWindow( view );
+			notifyListeners( view );
 			return view;
 		}
 		return null;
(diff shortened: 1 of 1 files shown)


tpietzsch:
The intention (notify about new MamutViewBdv) is fine.

Please use `Listeners.List<BdvViewCreatedListener>` for the member storing the listeners list and `Listeners<BdvViewCreatedListener>` for returning a `Listeners` handle where listeners can be un/registered (There should be one `bdvViewCreatedListeners()` method instead of methods for adding/removing).

The field should not simply be named `listeners`. I can already see the PR for adding `TrackSchemeViewCreatedListener`, and then that would be not specific enough.

See for example `FeatureTable` for the pattern to use
https://github.com/mastodon-sc/mastodon/blob/master/src/main/java/org/mastodon/feature/ui/FeatureTable.java#L141
https://github.com/mastodon-sc/mastodon/blob/master/src/main/java/org/mastodon/feature/ui/FeatureTable.java#L186
https://github.com/mastodon-sc/mastodon/blob/master/src/main/java/org/mastodon/feature/ui/FeatureTable.java#L299-L302

tpietzsch:
👍

ksugar:
Could you reopen this PR?  (or should I create a new PR)
I noticed that `bdvCreatedListeners` should be reset in `setAppModel()` instead of the constructor. With this change, `bdvCreatedListeners` will no longer accumulate every time you load/create a project.
https://github.com/elephant-track/mastodon/commit/d2348e7d4f83a4764f03c4b44abf073f4d0c701d

tpietzsch:
It seems wrong to clear registered listeners in `setAppModel()`.
Can you explain your problem in more detail? Probably there is a better way to solve it.

ksugar:
Thanks. I figured out that I needed to remove the listener on close of a BDV window.
The following change in my plugin solved the problem. `bdvCreatedListeners` does not need to be reset in `setAppModel()`.
https://github.com/elephant-track/elephant-client/commit/bc494df6e3237af33fe7a703fd58e20ff16b5379
The problem was that the previous listener would remain after loading/creating a project, but the above change solved that problem.

ksugar:
The above change was also wrong as you pointed out here https://github.com/elephant-track/elephant-client/commit/bc494df6e3237af33fe7a703fd58e20ff16b5379#r52880240
I hope I get the point, this time.
https://github.com/elephant-track/elephant-client/commit/ca6944bdb3c41a9b0d9fb271e6c271d7d274f904
Thank you for your kind instructions.

PR #131, opened 2021-06-03
Expose overlayCotextWrapper in BdvContextProvider

ksugar:
This PR exposes `overlayContextWrapper` in `BdvContextProvider`.

I made this change to get visible vertices in a BDV.

Example of usage:
https://github.com/elephant-track/elephant-client/blob/main/src/main/java/org/elephant/actions/mixins/WindowManagerMixin.java#L63-L74

```
final MamutViewBdv bdv = getBdvWindows().get( 0 );
@SuppressWarnings( "unchecked" )
final OverlayContextWrapper< Spot, Link > overlayContextWrapper = ( ( BdvContextProvider< Spot, Link > ) bdv.getContextProvider() ).getOverlayContextWrapper();
final Iterable< Spot > vertices = overlayContextWrapper.getInsideVertices( timepoint );
```

### DIFF
--- a/src/main/java/org/mastodon/mamut/MamutViewBdv.java
+++ b/src/main/java/org/mastodon/mamut/MamutViewBdv.java
@@ -353,4 +353,14 @@ public class MamutViewBdv extends MamutView< OverlayGraphWrapper< Spot, Link >, 
 	{
 		return coloringModel;
 	}
+
+	/**
+	 * Returns visible vertices in this BDV view.
+	 * 
+	 * @return {@code Iterable} of {@link Spot}
+	 */
+	public Iterable< Spot > getInsideVertices( final int timepoint )
+	{
+		return contextProvider.getOverlayContextWrapper().getInsideVertices( timepoint );
+	}
 }
--- a/src/main/java/org/mastodon/views/bdv/BdvContextProvider.java
+++ b/src/main/java/org/mastodon/views/bdv/BdvContextProvider.java
@@ -69,6 +69,8 @@ public class BdvContextProvider< V extends Vertex< E >, E extends Edge< V > >
 
 	private Context< V > context;
 
+	private final OverlayContextWrapper< V, E > overlayContextWrapper;
+
 	public BdvContextProvider( final String name,
 			final OverlayGraph< OverlayVertexWrapper< V, E >, OverlayEdgeWrapper< V, E > > overlayGraph,
 			final OverlayGraphRenderer< OverlayVertexWrapper< V, E >, OverlayEdgeWrapper< V, E > > renderer )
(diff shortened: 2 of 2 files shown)


tinevez:
Hello.

I have some comments.

I see you are trying to get what spots are visible in a BDV view right?
So you must go directly via the view object itself, not through the window manager. 
Can you try to remake this PR using this?

ksugar:
Hello, thank you for your review!
I will work on it.

ksugar:
I have updated the method to make it possible to call `getInsideVertices( timepoint )` from `MamutViewBdv`.

```
final MamutViewBdv bdv = pluginAppModel.getWindowManager().getBdvWindows().get(0);
final Iterable< Spot > vertices = bdv.getInsideVertices( timepoint );
```

> So you must go directly via the view object itself, not through the window manager.

I still need to get the view object through the window manager. How can I get it in a way that is not via the window manager?

tinevez:
I tried to walk around the code but could not find a shorter way to get what spots are displayed, respecting the current `RenderSettings`. The context ultimately defer to the renderer to know what spots are painted, so that is probably a correct way to do so. 

@tpietzsch What do you think of this? The goal is to make a public method that gives what spots are currently painted  in a specified view.

tinevez:
Hello @ksugar 

I am thinking more on your request, and how to make it more Mastodonic. 

Basically you have one 'thing' that needs to know what spots are displayed in a view. 
What about this 'thing' would directly subscribe as a listener to context changed? This is how we do it in TrackScheme and Tables, and this is how it is done in Mastodon.

Would that work for your application? Which is?

ksugar:
@tinevez 

Thank you for your feedback.
What I want to do is to get the visible spots in the selected BDV window. Any way is fine, as long as it achieves this goal.

I understood that this should be done via a listener for context changes.
In the [PR#129](https://github.com/mastodon-sc/mastodon/pull/129), I implemented a listener class for getting notifications when a new BDV window is created. Using it, I will be able to subscribe my plugin instance as a context listener.
If everything works, what I need will be only the [PR#129](https://github.com/mastodon-sc/mastodon/pull/129).

```
pluginAppModel.getWindowManager().addBdvViewCreatedListner( new BdvViewCreatedListener()
{

	@Override
	public void bdvViewCreated( MamutViewBdv bdv )
	{
		bdv.getContextProvider().listeners().add(MyPluginClass.this);
	}
} );
```

ksugar:
@tinevez
Following your suggestion, I could manage to get the vertices in a specific BDV window.
https://github.com/elephant-track/elephant-client/blob/dev/src/main/java/org/elephant/actions/BdvContextService.java
I will close this PR as I no longer need it.
Thank you for your feedback and suggestion.

PR #132, opened 2021-06-03
Add issort flag for Eigenvalue calculation

ksugar:
This PR adds an option to not perform sorting in the eigenvalue calculation.

In some cases, the sorting operations are unexpected.


tinevez:
Hello.

Just to sure: they are not sorted by default?

Can you add more to the doc:
- and state in what order they will be sorted, 
- on their value or on their absolute value?
- and stress that the vectors will be sorted as well, by the eigenvalues also.

ksugar:
Hello.

> Just to sure: they are not sorted by default?

They are **sorted** by default, which is the same behavior as before (I tried to keep the backward compatibility).

I did not modify the original sorting algorithm, which works as follows:

- eigenvalues (non-absolute) are sorted in ascending order accompanied by its corresponding eigenvectors

I will explicitly write it in the doc comments.

PR #133, opened 2021-06-03
Explicitly define DEFALULT_POINT_FADE_DEPTH

ksugar:
In `mastodon.views.bdv.overlay.RenderSettings`, the default value for `pointFadeDepth` was not set, while in `mastodon.views.bdv.overlay.ui.RenderSettingsIO.ConstructRenderSettings`, the default value was set as `DEFAULT_ELLIPSOID_FADE_DEPTH`.
In this PR, `DEFAULT_POINT_FADE_DEPTH` is defined as `0.0` and used in both classes.


tinevez:
All good for me.

PR #136, opened 2021-06-27
Add an option to fill spots in rendering

ksugar:
This PR allows users to set whether spots are filled or not from the settings dialog.
This option is helpful for visualization.
If this option is set, the spots will be drawn filled with the colour selected in `[View > Coloring]`, and the outline of the spots will be drawn in black.


tinevez:
Merged indirectly.

PR #137, opened 2021-07-02
Avoid deadlock in trackscheme

ksugar:
This PR is for avoiding deadlock in trackscheme.

### Issue
Opening the TrackScheme window while programmatically generating spots or links can cause a deadlock.

### Possible causes
- `graphRebuiilt()` might require writeLock because it affects Vertex and Edge models
- `JScrollBar.setValues()` inside readLock can cause a deadlock


tinevez:
How to reproduce the problem this PR fixes:
- Open a project (relatively huge one is better)
- Run [Plugins > Tracking > Detection...] with a low threshold value to generate a lot of spots (so that we can open the trackscheme during processing)
- Open a trackscheme

See below for an example of the problem.
https://files.gitter.im/60095ae2d73408ce4ff95c1a/Ehsi/TrackSchemeDeadLock.gif

PR #140, opened 2021-07-22
Make setting dialog scrollable (to dev)

ksugar:
(This PR is a duplicate of the PR #134 )

This PR makes the setting dialog scrollable.
It is helpful if the settings dialog is scrollable when there are many contents.


tinevez:
I don't like that the border of the panel is now visible: 
Screenshot 2021-07-22 at 12 00 46
I will fix that.

tinevez:
Done!

PR #144, opened 2022-02-17
added drag-and-drop handler for Fiji

xulman:
... with :heart: 

Pls, check that startMastodon() is starting it the expected way...

V.


tinevez:
Cool!
Can we make suggestions?

PR #147, opened 2022-04-01
Tiny fix, NullPointerException in FeatureComputationPanel/Model

maarzt:
I get a NullPointerExcpetion when I execute Mastodon.main(...) from IntelliJ.
The exception is thrown at this line: https://github.com/mastodon-sc/mastodon/blob/2f1572c2d9715371b6e59974546c32bd86dd13d5/src/main/java/org/mastodon/feature/ui/FeatureComputationPanel.java#L193

The reason is FeatureComputationModel.getFeatureSpec(...). It returns null if no feature is installed. It's probably better for this method to return a empty list rather the the null value, if there are no features found.

@tinevez You probably didn't came across this problem, because you the features installed in your project.


maarzt:
That was fast :+1: :heart:

PR #162, opened 2022-05-17
Branch graph and its applications in the Mastodon Mamut app.

tinevez:
# Branch graph and its applications in the Mastodon Mamut app.

During a recent discussion, Vlado and Mathias triggered in me the need to work again on the branch-graph framework in Mastodon. It resulted in this PR, documented below. Like a few before, the text is rather long and maybe over detailed. I hope to recycle parts of it for the documentation. If you want to get through it, it is probably best to print it and read it while stuck somewhere (12 pages when printed at a decent zoom level).

This PR does not amount to key changes in the core code, but _add_ code related to the branch-graph. 

The branch-graph is important I think for two main reasons:

- Views that are based on the branch-graph have a good performance, because the branch-graph is sparse compared to the core-graph it is based on. 
- Biology applications need the branch-graph. There are many important features that are defined on the branch-graph, such as cell-cycle length, cell excursion, cell hierarchy, etc. The branch-graph offers a convenient and sensible basis to compute and reason on them.

## Some history.

We started working on this 6 years ago I think during one of my autumn visit to Dresden. You proposed a first approach of a branch-graph class that is listening to the core graph and modifies itself following incremental changes. The changes are notified via the `GraphListener` interface, that tells what edge and vertex are added/removed in the core graph.

This initial work is present in the `mastodon-graph`, here:
https://github.com/mastodon-sc/mastodon-graph/tree/master/src/main/java/org/mastodon/graph/branch 

My recent (~1 month) efforts have been focused on finishing this one and implementing it as a useful, user-oriented set of features in the main Mastodon app (Mamut).

## Branch-graph implementations.

Despite my best efforts I could not implement a robust branch-graph that is listening to incremental changes. I went as far as I could but I could not make it perfect, and got quite depressed about it at some point. In brief, the incremental-changes version works in test cases, and in most cases in the Mamut app, but fails when the user does a big batch of changes (such as deleting half of the whole core graph). When I investigated it turns out that some of the map becomes invalid, but if I try to fix them other bugs appear. 

My crude solution was to rely on an implementation that does not listen to incremental changes, with another class. There are now two abstract branch-graph implementations. You can find them in branch `fix-branch-graph`, and there is an associated PR:

https://github.com/mastodon-sc/mastodon-graph/pull/8 

### `BranchGraphImp`

https://github.com/mastodon-sc/mastodon-graph/blob/fix-branch-graph/src/main/java/org/mastodon/graph/branch/BranchGraphImp.java

The `BranchGraphImp` class does not implement incremental changes. The branch-graph is only rebuilt upon calls of the `graphRebuilt()` method. The methods `edgeAdded(Edge)`, `edgeRemoved(Edge)`, `vertexAdded(Vertex)` and `vertexRemoved(Vertex)` do not do anything.

As a side effect, this implementation does not support 'rings'. If a connected component in the core-graph looks like this:

```
	 A -> B -> C -> D -> A
```

(edge direction is important), it won't be discovered in the branch-graph,which requires 'roots' to be built. (Roots are
vertices with no incoming edges.) 'Diamonds' connected components:

```
	A -> B -> D
and
	A -> C -> D
```

are supported.

### `BranchGraphImpIncremental`

https://github.com/mastodon-sc/mastodon-graph/blob/fix-branch-graph/src/main/java/org/mastodon/graph/branch/BranchGraphImpIncremental.java 

The `BranchGraphImpIncremental` class supports incremental changes. But there are still bugs when making batch changes in the core-graph. They are probably linked to not locking the core-graph and not cleaning the maps. This class should not be used without uttermost cautious, and could do with a revamp.

There are JUnit tests for both implementations.

There is a lot of common codes between the two implementations but we cannot subclass one with the other because we need to specifically call some methods in the superclass.

This is a crude way of casting away a piece of code I do not trust.

## In the Mastodon main app (Mamut).

We rely on the non-incremental version. The branch-graph is regenerated "on demand" via a button on one of the view or prior to computing features. On the Mette dataset I played with, the times it take to rebuild it is very small (210k vertices, 3k branches).

### The branch-graph concrete implementation.

Nothing special: It is a `ModelBranchGraph ` inheriting from the abstract `BranchGraphImp` class mentioned above, with a `BranchSpot` as vertex and `BranchLink` as edge.

- https://github.com/mastodon-sc/mastodon/blob/branch-graph-v2/src/main/java/org/mastodon/mamut/model/branch/BranchSpot.java
- https://github.com/mastodon-sc/mastodon/blob/branch-graph-v2/src/main/java/org/mastodon/mamut/model/branch/BranchLink.java

The `BranchSpot` exposes the properties of the `Spot` it is linked to.

### Changes in the `Model` class.

The model now has two new fields:

```java
public Model( final String spaceUnits, final String timeUnits )
{
...
	branchGraph = new ModelBranchGraph( modelGraph, initialCapacity );
	branchIndex = new SpatioTemporalIndexImp<>( branchGraph, branchGraph.getGraphIdBimap().vertexIdBimap();
...
```

As noted above, the `graphRebuilt()` method of `branchGraph` has to be 'manually' called for the branch-graph to be in sync. There is a small class with a listener that lets the user do that:

https://github.com/mastodon-sc/mastodon/blob/branch-graph-v2/src/main/java/org/mastodon/mamut/model/branch/BranchGraphSynchronizer.java

It is a field of the `MamutAppModel` and is used by UI components to 

- report whether or not the branch-graph is in sync,
- regenerate the branch graph.

#### A new hierarchy for branch views.

I wanted to reuse the view window hierarchy we have for the core-graph view (https://github.com/mastodon-sc/mastodon/blob/branch-graph-v2/src/main/java/org/mastodon/app/MastodonView.java etc) but could not find a way to reconcile it with to generate a view of the branch-graph. The main disagreement stands in that these classes are typed for a certain `Graph`, that is itself typed for `AbstractSpot` as vertex. The `BranchSpot` does not inherit from it; it merely returns the `Spot` it is linked to.

So I made a new hierarchy for the branch-graph views, trying to find a common interface for their common methods.

https://github.com/mastodon-sc/mastodon/blob/branch-graph-v2/src/main/java/org/mastodon/mamut/MamutBranchView.java

This is where the `SelectionModel`, `FocusModel`, etc. of the core-graph are _adapted_ to the branch-graph. The branch-graph does not have a special selection model, but adapts the one existing for the core-graph. This allows having common events. For instance, selecting a branch in a branch-view selects all the spots and links linked to the branch. A branch is _shown_ to be tagged with a certain tag iif all the spots and links of the branch have the same tag. Etc. I detail below how each model component is adapted. The mechanisms of adaptation are detailed below.

The two next views inherit from this common class. 

#### The branch BDV view.

It has nothing special. It just paints the branch-graph like any other graph. It could be used as a source for spatial `Context` but this PR does not make use of it for now.

#### The branch TrackScheme view.

I think this is the one that is interesting for our users. From this kind of TrackScheme:

Mastodon_branch-graph_01

the branch TrackScheme looks like this:

Mastodon_branch-graph_03

(Notice that the trees in the branch TrackScheme are laid-out  in the same order that of the main TrackScheme. Pure luck.) 

And the branch edges have square elbow shapes:

Mastodon_branch-graph_02

Note that there is a "Regen branch graph" button, based on the synchronizer I mentioned above. When it is displayed with a green dot, it means that the branch graph is in sync. When it is displayed with a cog, it means the branch-graph is out of sync.

Four things were needed to have this display of the branch-graph:

1. The `TrackSchemeGraph` must listen to the branch-graph in a clever way. The branch-graph reports that it changed with the `GraphChangeListener` interface, but the events need to trigger a call to `graphRebuilt()` in the `TrackSchemeGraph`. If not, TrackScheme will make a layout with duplicated trees. This happens because the `roots` field is not cleared by the `graphChanged()` method, as expected (guess how I found out). This is why in the view constructor we have:

```java
// Forward graph change event to a graphRebuilt event for branch graph.
viewGraph.graphChangeListeners().remove( frame.getTrackschemePanel() );
viewGraph.graphChangeListeners().add( () -> {
  viewGraph.getLock().writeLock().lock();
  try
  {
    viewGraph.graphRebuilt();
  }
  finally
  {
    viewGraph.getLock().writeLock().unlock();
  }
  frame.getTrackschemePanel().graphChanged();
} );
```

We loose the nice animation like the one we have in the core-graph TrackScheme when an incremental change happens, but this is life.

2. I wanted to paint the edges in a custom way, with the square elbow. So I made a new `TrackSchemeOverlayFactory`, that uses the new `PaintBranchGraph` class:

https://github.com/mastodon-sc/mastodon/blob/branch-graph-v2/src/main/java/org/mastodon/views/trackscheme/display/PaintBranchGraph.java

I also made the default stroke of the branches a bit thicker, to maximise the user experience.

3. Normally in TrackScheme, we only paint the edges of the vertices that are visible in the view window. This is fine when the graph is dense, but absolutely not for the branch-graph, which is very sparse. At some point in the past I looked if we could have the edges added to a special structure such as B-tree, but gave up. 

Here the simple solution I use is to paint the vertices irrespective of the time-point. So when we are displaying a box `x, y, w, h` in TrackScheme, I actually generate the screen objects for all the vertices that are in `x, 0, w, height`. 

You can 'feel' it a bit: The core-graph TrackScheme has a better performance when fully zoomed in than the branch-graph TrackScheme at this zoom level. I think it is good enough for now. ♨️

4. The `ScreenRanges` objects are used when the vertices become very densely packed on the screen. But again they expect the vertices to be dense with one vertex every frame. I did not find a good way to reconcile this with the branch graph, where all branches are possibly of different lengths and are made by only 2 vertices spaced in time by a lot. 

So the branch-graph TrackScheme do not use them. This results in us painting the whole branch-graph when fully zoomed out. ♨️

#### Integration in the main UI.

I did not want to disturb too much the UI with the branch-graph, that might not be used by everyone. So the two new views above can be accessed by shift-clicking on the bds or trackscheme buttons. Or via the menu items in the Window menu.

#### The new data table.

We also want to display the branch-graph in the data table, because there are numerical features on branches that are interested in. Rather than making an additional, separate view that display an edge table and a vertex table for a branch graph, I reworked the table view so that it can incorporate the two. It looks like this:

Mastodon_branch-graph_04

There are 4 tabs, 2 for the core-graph, 2 for the branch-graph. 

Remaking the table view forced me to think at how a API consumer would use this, so I made a cool table builder that lets configure what you want and don't want in the table. For instance, the view above is built with a call to the table builder that looks like this:

```java
		// Create tables.
		final TableViewFrameBuilder builder = new TableViewFrameBuilder();
		final MyTableViewFrame frame = builder
				.groupHandle( groupHandle )
				.undo( model )
				.addGraph( model.getGraph() )
					.selectionModel( selectionModel )
					.highlightModel( highlightModel )
					.focusModel( focusModel )
					.featureModel( featureModel )
					.tagSetModel( tagSetModel )
					.navigationHandler( navigationHandler )
					.coloring( coloringAdapter )
					.vertexLabelGetter( s -> s.getLabel() )
					.vertexLabelSetter( ( s, label ) -> s.setLabel( label ) )
					.listenToContext( true )
					.selectionTable( selectionTable )
					.done()
				.addGraph( model.getBranchGraph() )
					.vertexLabelGetter( s -> s.getLabel() )
					.vertexLabelSetter( ( s, label ) -> s.setLabel( label ) )
					.featureModel( featureModel )
					.tagSetModel( branchTagSetModel( appModel ) )
					.selectionModel( branchSelectionModel( appModel ) )
					.highlightModel( branchHighlightModel( appModel ) )
					.coloring( branchColoringAdapter )
					.focusModel( branchFocusfocusModel( appModel ) )
					.navigationHandler( branchGraphNavigation( appModel, navigationHandler ) )
					.done()
				.title( selectionTable ? "Selection table" : "Data table" )
				.get();
```

You add graphs you want to display in the window by calling `.addGraph()` then the builder switches to another class that lets you specify stuff like the selection model, how to display object labels, whether to listen to a context, etc. If you don't specify something, its functionality won't be present, but it will not crash. Then you call `.done()` to return to the main builder, and finish building the window by calling `.get();` which returns a ready-made `ViewFrame` properly registered with an `onClose()` mechanism to deregister listeners.

### Model components.

As stated above, the branch-graph does not have a special selection, a highlight, etc. It simply relies on adapters that map the ones from the core-graph. 

Now there is a big danger, because we do not use the incremental-changes version of the branch-graph, so the two graphs might be out of sync while they are displayed. The branch-graph might link to spots and links that have been deleted. I did not try to fix this. I circumvent issues by implementing a policy that says: "if you cannot find the object linked to the one you are dealing with, please don't crash, but just give up."

This is implemented in the `AbstractBranchGraphAdapter` class, that is the mother class for all the branch-graph model component adapters. It has two methods `isValid( final V v )` and  `isValid( final E e )` that returns true only if mapping from a branch-graph object to a core-graph object can be found and is valid. The concrete adapters deal with invalid objects in a manner that makes sense for the model they adapt. 

#### Branch selection-model.

- Selecting a branch link selects all the spots and links that are linked to it.
- Selecting a branch spot selects the spot it is linked to. And conversely.
- Conversely, a branch link will be shown as selected if and only if all the spots and links that are linked to it are selected. 

#### Branch highlight.

- Highlighting a branch spot will highlight the spot it is linked to. And conversely. 
- Highlighting a branch link will highlight the link it is linked to. (This is the first outgoing link of the branch.) And conversely. 

#### Branch focus.

- Focusing on a branch spot will focus the spot it is linked to. And conversely. 

#### Group-handle and navigation.

Nothing special for the group-handles. They use the same model that for the other views.

- Navigating to a branch spot will navigate to the spot it is linked to. And conversely. 
- Navigating to a branch link will navigate to the link it is linked to. And conversely. 

#### Tag-set for branch items.

Again, the branch-graph does not have a separate model for tag-sets, but adapts the tag-set model of the core graph. This is done in a similar manner than for the selection:

- Tagging a branch spot selects tags spot it is linked to. And conversely.
- Tagging a branch link tags all the spots and links that are linked to it.
- Conversely, a branch link will be shown as tagged with a certain tag if and only if all the spots and links that are linked to it are tagged with the same tag. 

This gives this kind of view:

Mastodon_branch-graph_05

Mastodon_branch-graph_06

But observe below for instance that the branch tagged in blue in the core-graph does not appear as blue in the branch graph becomes some links are not tagged:

Mastodon_branch-graph_07

Mastodon_branch-graph_08

#### Branch-graph features.

The Mamut feature computation framework is augmented so that it can deal with feature computers for branch items, and provide the parameters they need for computation. A branch-graph feature can depend on a core-graph feature and conversely. Their computation is configured in the same UI:

Mastodon_branch-graph_09

All the features, regardless of their type, are stored in the same `FeatureModel` instance, which does not mind. The table views discussed above know how to retrieve the right features for the data they display.

##### Coloring models.

All the features, for the branch-graph and the core-graph, are stored in the same feature model instance. But we still need to generate colors in a separate manner. At the same time I did not want to multiply the number of classes by 2 every time there is a functionality related to branch-graph to add. 

So I added new modes in the `VertexColorMode` and  `EdgeColorMode` enum classes, and made two `ColoringModel`s. There is one for the core-graph views (`ColoringModelMain`), that can interpret all the coloring modes listed below, using a core-graph feature or a branch-graph feature. And there is another one for branch-graph views (`ColoringModelBranchGraph`) for branch-graph views using only branch-graph features. We now support the following color modes:

For the core-graph:

- For the `Spot`s of the core-graph, there is now 8 coloring modes, including 4 new ones using the branch-graph.
  - Coloring based on a spot feature. `VERTEX`
  - Color based on a link feature, using the spot sole incoming link (`INCOMING_EDGE`) or its sole outgoing link (`OUTGOING_EDGE`). If there are 0 or more than 1 incoming, respectively 1 outgoing, link, we give up and return a default color.
  - No coloring, the default. `NONE`.
  - Color based on a branch-spot feature, looking _up_ (backward in time in the same branch) or _down_ (forward in time) for the branch-spot linked to this spot. `BRANCH_VERTEX_UP` and `BRANCH_VERTEX_DOWN`.
  - Color based on branch-link feature, looking for the branch-link linked to the spot sole incoming link (`INCOMING_BRANCH_EDGE`) or its sole outgoing link (`OUTGOING_BRANCH_EDGE`). If there are 0 or more than 1 incoming, respectively 1 outgoing, link, we give up and return a default color.
- For the `Link`s of the core-graph, there is now 9 coloring modes, including 5 new ones using the branch-graph.
  - Coloring based on a link feature. `EDGE`
  - Coloring based on a spot feature, using either the source spot or the target spot of the link. `SOURCE_VERTEX` and `TARGET_SPOT`.
  - No coloring, the default. `NONE`.
  - Coloring based on a branch-spot feature, looking respectively _up_ (backward in time) or _down_ (forward in time) for the branch-spot linked to their source spot, respectively target spot. `SOURCE_BRANCH_VERTEX_UP`, `TARGET_BRANCH_VERTEX_UP`, `SOURCE_BRANCH_VERTEX_DOWN` and `TARGET_BRANCH_VERTEX_DOWN`.
  - Coloring based on a branch-link feature, looking for the branch-link of the branch the link is in. `BRANCH_EDGE`.

The branch graph has the same color modes that the core graph before this PR. They always relate to branch-spot feature or a branch-link feature. There are no way to color a branch-graph object with a core-graph feature. (I could not think of a sensible way to do that. Taking the mean on the fly? Would not it be simpler to define a branch-spot feature that computes the mean of a core-graph feature, but not on the fly?)

- For the `BranchSpot`s:
  - Coloring based on a branch-spot feature. `VERTEX`, `BRANCH_VERTEX_UP`, and `BRANCH_VERTEX_DOWN`. 
  - Color based on a link feature, using the spot sole incoming link (`INCOMING_EDGE` and `INCOMING_BRANCH_EDGE`) or its sole outgoing link (`OUTGOING_EDGE`, `OUTGOING_BRANCH_EDGE`). If there are 0 or more than 1 incoming, respectively 1 outgoing, link, we give up and return a default color.
  - No coloring, the default. `NONE`.

- The `BranchLink`s:
  - Coloring based on a link feature. `EDGE`
  - Coloring based on a spot feature, using either the source spot or the target spot of the link. `SOURCE_VERTEX`, `SOURCE_BRANCH_VERTEX_DOWN`, `SOURCE_BRANCH_VERTEX_UP`, and respectively`TARGET_SPOT`, `TARGET_BRANCH_VERTEX_DOWN`, `TARGET_BRANCH_VERTEX_UP`.
  - No coloring, the default. `NONE`.

##### The Feature color mode collection.

The Mamut app we use manages a collection of feature color modes, built on the vertex and edge modes discussed above. Normally we should manage a second collection of coloring modes for the branch-graph views. But I also wanted to avoid multiplying the number of fields in the model. 

So there is still only one collection of feature color modes, common to the two coloring models discussed above. The coloring models simply filter them also based on the vertex and edge modes. For instance, the coloring model for branch-graph views will tag as invalid a feature color mode that has `VERTEX` for the vertex mode, because it is based on a spot feature. This way the view menu shows all the modes, but only the ones supported in this view are enabled:

Mastodon_branch-graph_11

Mastodon_branch-graph_10

The data table view is special, as it displays both the core-graph and the branch-graph. So there are two coloring menus:

Mastodon_branch-graph_12

##### The Feature color mode creation UI.

Since the feature-model is happy to welcome within in features that apply to different object types, I tried to reused the UI we have to specify feature color modes. Again, the goal was to avoid multiplying the number of UI windows we show to the user. 

The vertex and edge mode selectors are now in a scroll pane to show all the modes, and the non selected ones are greyed for clarity. I also tried to improve the labels, so that we show 'Spot' and 'Link' in the Mamut app, instead of 'Vertex' and 'Edge', again to make it simpler and clearer to the user:

Mastodon_branch-graph_13

## After this PR.

Questions about what is possible next:

- Shall we serialize the branch-graph? So that specialized readers (MATLAB / Python) can de-serialize it?
- Make a view for a hierarchy tree, based on the branch-graph. That is: a view of the data like in TrackScheme, but where the Y position is not time, but depth in the tree, or generation. Mette asked for this to Vlado and Mathias I think, to make sense of her lineages in an easier way.
- Consider coloring branches based on Spot or Link features? Taking the mean on the fly? Would not it be simpler to define a branch-spot feature that computes the mean of a core-graph feature, but not on the fly?


maarzt:
Oh, it's merged already. Rightfully so. It works really nicely. I noticed some minor things thought, which we can patch later:
* Using the left and right key when navigating the branch graph feels wrong. The trackscheme window jumps to far away points.
* The diameter of nodes shown in the barnch graph trackscheme feels wrong / random.

tinevez:
Hello @maarzt 

Yep it's in the `dev` now. Nothing prevents you from branching to modify the things you want.

1. Yes, and I am afraid it's normal: it jumps to the next spot in the same frame. But on the branch-graph there are few spots.

2. Yes agreed. I think the way we compute the distance to the closest spot is not well suited to the situation where we have few spots spaced by many time-points.

Do you will like working on those?

PR #168, opened 2022-05-27
Ask the user to save the project before closing the main window v2.

tinevez:
When closing the main window, this dialog is shown. But only if there have been changes since the last save.

image


tinevez:
Supersedes https://github.com/mastodon-sc/mastodon/pull/51

Issue #172, opened 2022-06-09
TrackScheme Hierarchy improvements


maarzt:
Improve the rendering of the hierarchy view, to make it easier to grasp to structure of the lineage:

- [x] Display the node labels, not inside the circle, but below or next to the circle.
      (The circle visualizing the nodes are usual not big enough to show the node label.)
- [x] Have wider lines (~2 or 3 pixels) connecting the nodes.
- [x] None curved lines. (I'm sorry! The curves are BEAUTIFUL. But they visually confuse me to much.)
- [x] Restrict the hierarchy view to only show one sub-tree of the branch graph. The biologists are often only interested in one particular cell, and how it divides. So the would love to focus on this one cell, and only show its lineage.

tinevez:
Quite some work :)
Do you feel like working on it?

maarzt:
@tinevez Yes, I'm indeed already working on some of these points.

tinevez:
All done!
The issue with the node size the branch TrackScheme is the subject of another issue, the resolving of which is postponed.

PR #174, opened 2022-06-15
Changes to the hierarchical TrackScheme

maarzt:
This PR changes the Hierarchical TrackScheme. The goal is to allow the users to get a better overview of the data. The rendering of the Hirarchical TrackScheme is changed slightly. And there are new menu items. That allow to only show a selected track or downward track in the Hierarchical TrackScheme.

This PR fixes https://github.com/mastodon-sc/mastodon/issues/172

**Changes to the  rendering of the Hierarchical TrackScheme:**
* Use edges with corners, like in the branch graph TrackScheme.
* Wider lines.
* All nodes have the same size, which only depends on the zoom factor.
* Put the labels below the nodes.

image

**Allow to zoom out further in the hierarchical TrackScheme:**
image

**Allow to show only a subset of the tracks in the hierarchical TrackScheme:**
image

**Refacoring of the LineageTreeLayout class**

I had a deeper look into the LineageTreeLayout class in order to implement the new functionality. The code was quite complicated so a refactored a lot of stuff such that I could understand it better:
* Make `LineageTreeLayout` an interface an move the implementation to `LineageTreeLayoutImp`
* Replace the boolean `paintLongEdges` setting in `TrackSchemeOptions` by using a `LineageTreeLayoutFactory`.
* Simplify the quote "impenetrable" `layoutX_iterative()` method in `LineageTreeLayout` by using a new `DepthFirstIteration` class.
* Move code for managing a Table of `TrackSchemeVertices` from `LineageTreeLayout` to `TrackSchemeVertexTable`.
* Reduce the code in `LineageTreeLayout` and `LongEdgesLineageTreeLayout` by adding the new methods `addScreenVertex` and `buildScreenColumns`

**New HierarchyLayout class**
I added a new HierarchyLayout class the only purpose is to hide the label of the root node. The label would otherwise be printed over an edge. And the label is anyway visible as the title of the column.


maarzt:
Some thoughts:
* Should the list of root nodes to be shown, be held in the `LongEdgesLineageTreeLayout`?
* We can have the same "Show only selected tracks" functionality for the normal TrackScheme view. It would actually be pretty simple. We only need to move some code from `MamutBranchViewTrackScheme` to `MamutViewTrackScheme` and from `LongEdgesLineageTreeLayout` to `LineageTreeLayoutImp`.

src/main/java/org/mastodon/views/trackscheme/display/PaintBranchGraph.java
```diff
@@ -43,7 +43,7 @@
 public class PaintBranchGraph extends PaintGraph
 {
 
-	private final Stroke edgeStroke = new BasicStroke( 1.5f );
+	protected Stroke edgeStroke = new BasicStroke( 1.5f );
```
tinevez, line 46: What about we define a stroke for 
- normal TrackScheme
- branch TrackScheme
- hierarchy TrackScheme
in the `TrackSchemeStyle`? 
Possibly we could later let the user edit them.

src/main/java/org/mastodon/views/trackscheme/display/PaintHierarchicalGraph.java
```diff
@@ -0,0 +1,156 @@
...
+				label = label.substring( 0, maxLabelLength - 2 ) + "...";
+
+			if ( !label.isEmpty() )
+			{
+				g2.setColor( labelColor );
+				final FontRenderContext frc = g2.getFontRenderContext();
+				final TextLayout layout = new TextLayout( label, style.getFont(), frc );
+				final Rectangle2D bounds = layout.getBounds();
+				final float tx = ( float ) ( x - bounds.getCenterX() );
+				final float ty = ( float ) ( y + bounds.getHeight() + 1.2 * spotRadius + STROKE_WIDTH );
```
tinevez, line 151: With this the round circle of the node is a bit empty and feels useless. 
Would you feel like make a display mode where the labels are painted in the nodes, but with nodes having a rounded rectangle shape? Its max width would be set by the column width of the branch child breadth.
Then we could make a flag or an enum that could let the user decide how they want to paint them.
What do you think?

src/main/java/org/mastodon/views/trackscheme/display/PaintHierarchicalGraph.java
```diff
@@ -0,0 +1,156 @@
...
+
+	@Override
+	protected void drawVertex( ScreenVertex vertex )
+	{
+		if(spotRadius > 3)
+			drawVertexFull( vertex );
+		else
+			drawVertexSimplified( vertex );
+	}
+
```
tinevez, line 96: Oh I miss my roundish corners :)
Any chance it could be back, not set by default, and again made configurable in the `TrackSchemeStyle` object?
What do you think?

tinevez (approved):
This is fantastic. You will find here comments that are suggestions. Check whether you feel like implementing them.

src/main/java/org/mastodon/views/trackscheme/display/ShowSelectedTracksActions.java
```diff
@@ -0,0 +1,213 @@
...
+import org.scijava.ui.behaviour.util.Actions;
+import org.scijava.ui.behaviour.util.RunnableAction;
+
+/**
+ * These actions allow to limit the view of the
+ * {@link org.mastodon.mamut.MamutBranchViewTrackScheme}
+ * and {@link org.mastodon.mamut.MamutBranchViewTrackSchemeHierarchy}
+ * to only show selected tracks or downward tracks.
+ */
+public class ShowSelectedTracksActions
```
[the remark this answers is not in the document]
maarzt: Yes, I just added this functionality for the normal TrackScheme.

src/main/java/org/mastodon/views/trackscheme/display/PaintHierarchicalGraph.java
maarzt: That's a good idea. Let's introduce an option for that.

I prefer the round corners aesthetically. But with the sharp corners I can focus better on the tree structure? Did you observe something similar?

tinevez: ..... no. 
I guess it is a matter of taste rather than Science, so it has its place in the settings editor.

src/main/java/org/mastodon/views/trackscheme/display/PaintHierarchicalGraph.java
tinevez: Hehe :) OK.

src/main/java/org/mastodon/mamut/MamutBranchViewTrackSchemeHierarchy.java
```diff
@@ -55,6 +58,7 @@ public MamutBranchViewTrackSchemeHierarchy( final MamutAppModel appModel, final
 				if ( t < minT )
 					minT = t;
 			}
+			maxT = Math.max( MINIMUM_NUMBER_OF_HIERARCHY_LEVELS_SCROLLABLE, maxT );
```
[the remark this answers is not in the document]
maarzt: This is indeed a hack to achieve the following. The user should be able to zoom out further in Y. Her is what it looks like to fully zoom out without this line:
image

With this line:
image

src/main/java/org/mastodon/util/DepthFirstIteration.java
```diff
@@ -0,0 +1,236 @@
+/*-
```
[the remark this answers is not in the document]
maarzt: OK, I checked. Unfortunately neither of the to classes `DepthFirstIteration` and `DepthFirstSearch` cannot be used to replace this DepthFirstIteration class. The `DepthFirstIterator` doesn't visit the nodes twice. The second visit is actually needed in the layoutX() method. The `DepthFisrtSearch` is based on a recursive function call, which cannot provide the required performance. `DepthFisrtSearch` will likely cause a StackOverflowException when used in the layoutX method.
Another important feature that neither of the implementations provide is the ability the exclude a subtree from the iteration.

Yes, the DepthFirstIteration can be added to the mastodon-graph repo, but we would need to rewrite the test.

src/main/java/org/mastodon/views/trackscheme/display/PaintBranchGraph.java
maarzt: Yes, it should be done as you suggest.

maarzt:
@tinevez Sorry, I could finish this today. And I won't have time to look at this before next week.

src/main/java/org/mastodon/mamut/MamutBranchViewTrackScheme.java
```diff
@@ -229,6 +240,15 @@ protected MamutBranchViewTrackScheme(
 
 		// Give focus to the display so that it can receive key presses immediately.
 		frame.getTrackschemePanel().getDisplay().requestFocusInWindow();
+
+		// Automatically regenerate branch graph, if it never hase been generated before
+		if(appModel.getModel().getBranchGraph().vertices().size() == 0)
```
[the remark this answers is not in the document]
tinevez: Yes they need to be recomputed. It is the same that for the core graph. 
When the user trigger a feature computation btw, the branch graph is regenerated automatically.

src/main/java/org/mastodon/util/DepthFirstIteration.java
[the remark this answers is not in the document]
tinevez: Ok I make an issue about that.

maarzt:
Hi @tinevez, I will start now to fix the remaining issues.

maarzt:
@tinevez Ok, I think it's ready now. Please review my last changes and merge it.

maarzt:
:partying_face:

PR #184, opened 2022-07-27
Menu and title fixes

xulman:
I have accidentally discovered:
Screenshot_20220727_151026

and while fixing it, I synced item names and camelCaseiness with the actual window Names


xulman:
APOLOGIES, did against wrong branch... obsolete, continue pls here #185

PR #186, opened 2022-08-02
Change the "edit label action" to rename all spots of a branch at once (Branch&Hierarchy TrackScheme)

maarzt:
Mastodon users usually would like to assign the same name to all the spots in a (BranchGraph) branch.
Both the **Branch TrackScheme** and **Hierarchy TrackScheme** could be very useful in this context.

The normal TrackScheme has a `EditFocusVertexLabelAction` for renaming individual spots.
This PR extend this class as a new `BranchTrackSchemeEditLabelAction`.
This `BranchTrackSchemeEditLabelAction` is used in the Branch- and Hierarchy TrackScheme to rename all the spots that belong to a branch at once.


src/main/java/org/mastodon/views/trackscheme/display/EditFocusVertexLabelAction.java
```diff
@@ -180,6 +180,12 @@ public void updateHeaderSize( final int width, final int height )
 		headerHeight = height;
 	}
 
+	protected void changeLabel( TrackSchemeVertex vertex, String label )
+	{
+		vertex.setLabel( label );
+		undoPointMarker.setUndoPoint();
```
tinevez, line 186: We probably should make an undo point only once, after the whole branch has been renamed. 
Here we store an undo point for every vertex being renamed.

src/main/java/org/mastodon/views/trackscheme/display/BranchTrackSchemeEditLabelAction.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	@Override
+	protected void changeLabel( TrackSchemeVertex vertex, String label )
+	{
+		vertex.setLabel( label );
+		BranchSpot branchSpot = trackSchemeGraph.getVertexMap().getLeft( vertex );
+		Spot vertexRef = graph.vertexRef();
+		Spot spot = branchGraph.getLinkedVertex( branchSpot, vertexRef );
+		while(spot.incomingEdges().size() == 1)
```
tinevez, line 52: I am not sure of this (on holidays, limited Eclipse access) but I think there is a branch iterator somewhere in the code. Could you check and if it exists reuse it?

src/main/java/org/mastodon/views/trackscheme/display/EditFocusVertexLabelAction.java
tinevez, line 186: We probably should make an undo point only once, after the whole branch has been renamed.
Here we store an undo point for every vertex being renamed.

tinevez (approved):

src/main/java/org/mastodon/views/trackscheme/display/EditFocusVertexLabelAction.java
maarzt: There a little confusion here going on. This method is not used to rename the branch labels. This is the method that is used to rename a single individual spot in the normal TrackScheme window, when pressing enter. Here we want an undo point after each individually renamed spot.

*(The source code is changed here to make it easier for me to extend this class for the Branch TrackScheme. Notice how the method is protected. It is overwritten in BranchTrackSchemeEditLabelAction.)*

src/main/java/org/mastodon/views/trackscheme/display/BranchTrackSchemeEditLabelAction.java
maarzt: I couldn't find any BranchIterator, or similar. An InverseDepthFirstIterator could maybe be used?

tinevez:
Hello @maarzt 
I made some small changes. Could you check them?

PR #193, opened 2022-09-15
Enable opening of dummy data

xulman:
Previously there was BasicImgLoader utlized in `DummySpimData` object.
But [some code around](https://github.com/mastodon-sc/mastodon/blob/adc7bddb4850510007b1a93004f33d04b7bb2340/src/main/java/org/mastodon/views/bdv/SharedBigDataViewerData.java#L392) (even [outside Mastodon](https://github.com/bigdataviewer/bigdataviewer-core/blob/9255efe8eb63a4b861ea7c6be283b556d31c0e7a/src/main/java/bdv/BigDataViewer.java#L277)) is casting to `ViewerImgLoader`.

To enable these pieces of code to function with our dummy data, we had to upgrade the dummy....

The implementation features single resolution level pyramid (I hope).

### DIFF
--- a/src/main/java/org/mastodon/util/DummySpimData.java
+++ b/src/main/java/org/mastodon/util/DummySpimData.java
@@ -34,11 +34,13 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import bdv.ViewerImgLoader;
+import bdv.ViewerSetupImgLoader;
+import bdv.cache.CacheControl;
 import bdv.spimdata.SequenceDescriptionMinimal;
 import bdv.spimdata.SpimDataMinimal;
 import bdv.util.ConstantRandomAccessible;
 import mpicbg.spim.data.generic.sequence.BasicImgLoader;
-import mpicbg.spim.data.generic.sequence.BasicSetupImgLoader;
 import mpicbg.spim.data.generic.sequence.BasicViewSetup;
 import mpicbg.spim.data.generic.sequence.ImgLoaderHint;
 import mpicbg.spim.data.registration.ViewRegistration;
@@ -49,8 +51,11 @@ import net.imglib2.Dimensions;
 import net.imglib2.FinalDimensions;
 import net.imglib2.FinalInterval;
 import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.Volatile;
 import net.imglib2.realtransform.AffineTransform3D;
+import net.imglib2.type.numeric.RealType;
 import net.imglib2.type.numeric.integer.UnsignedShortType;
+import net.imglib2.type.volatiles.VolatileRealType;
 import net.imglib2.view.Views;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/mastodon/util/DummySpimData.java
```diff
@@ -139,22 +144,65 @@ private static double get( final String[] parts, final String key, final double
...
 			{
+				@Override
+				public RandomAccessibleInterval<T> getImage(int timepointId, int level, ImgLoaderHint... hints) {
+					return img;
+				}
+
+				final double[][] resolution = new double[1][3];
+				final AffineTransform3D[] transform = new AffineTransform3D[1];
+				{
+					resolution[0][0] = 1.0; //TODO should be taken from 'calib' from our only-one caller code...
```
xulman, line 174: I'm not sure how much important is to fix this TODO given that we talk here about DUMMY data...

if we were to make it crystal clear, we would need to modify the c'tor (easy, there is only one caller of it) to bring in ["calibration" and "transformation"](https://github.com/mastodon-sc/mastodon/blob/adc7bddb4850510007b1a93004f33d04b7bb2340/src/main/java/org/mastodon/util/DummySpimData.java#L110-L120)

tinevez:
Maybe this PR could be the opportunity to improve opening a mastodon file when the image data is missing. For now several things could be improved:

- [ ] The user gets a convoluted error message in the IJ console. We could give them a nice warning window, likethe ones we use when opening from ImageJ ImagePlus.
image

- [ ] The BDV appears empty of tracks when you open it. The user must dezoom or pan or zoom to make them appear. Maybe because of an incorrect bounding box determination at start?
image

- [ ] Same thing for time: TrackScheme opens with a max timepoint of 9 and cannot zoom out. It does not determine max timepoint from the data:
image

xulman:
yes, agree to all

btw, I've revamped the File -> Fix Image Path dialog in mastodon-tomancak repo recently;
on the master in m.-tomancak repo it looks now like this:  
Screenshot_20220919_115401

In the dummy sub dialog I could maybe pre-load default values from the `Graph`, and we could try to open this thing whenever loading of project fails?

what do you think? @tinevez @tpietzsch @maarzt

xulman:
so, to the latter two points of JY above:
the dummy data is some default `Interval`, likely small.. and thus off compated to where the spots actually are, thus the BDV view looks at nothing

similarly to the timpoints range, the dummy data by default creates only 10 timepoint, which is why TS win was "trimmed" (also BDV window should have time slider only for 10 TPs)

while i didn't care too much about the spatial extent of the dummy data (it's empty, why would I even cared to open BDV...), I cared about the temporal limit and that actually motivated me to introduce the "I want dummy data" button in which *especially* the time span can be adjusted

xulman:
actually was now tracing how default dummy data is created:
- https://github.com/mastodon-sc/mastodon/blob/9b786081652a2975bf84948281cdc2db8bf75b79/src/main/java/org/mastodon/mamut/ProjectManager.java#L600
- [if we have some `.xml` data pointer, we open it directly here](https://github.com/mastodon-sc/mastodon/blob/9b786081652a2975bf84948281cdc2db8bf75b79/src/main/java/org/mastodon/mamut/ProjectManager.java#L639)
- [it then assumes it is a BDV data](https://github.com/mastodon-sc/mastodon/blob/9b786081652a2975bf84948281cdc2db8bf75b79/src/main/java/org/mastodon/mamut/ProjectManager.java#L925)
- [after failing opening the data, it inspects...](https://github.com/mastodon-sc/mastodon/blob/9b786081652a2975bf84948281cdc2db8bf75b79/src/main/java/org/mastodon/views/bdv/SharedBigDataViewerData.java#L384)
- [...with defaults](https://github.com/mastodon-sc/mastodon/blob/9b786081652a2975bf84948281cdc2db8bf75b79/src/main/java/org/mastodon/util/DatasetInfoParser.java#L104) which is where the 10 TPs are set

hmm.. it's a puzzle... when the `.xml` file is missing, it does not even reach the `inspect()` code...

xulman:
* rebased against the latest `origin/dev` brach
* added one commit that attempts to clean up a bit the console reporting... the main contribution is basically only the replacing of `printStackTrace()` with `getMessage()`... no big, but looks less messy

xulman:
so, this is what it looks like:
Screenshot_20220919_163028

now, since it actually opens something, two consequences:
- needs to open something to be able to fix its dataset path
- the launcher window is gone and makes no sense putting info/warn/error msg into it

@tinevez I would say this could close the item 1 "user gets convoluted message", what do you think?

xulman:
Screenshot_20220919_171347

@tinevez I would say this could close the item 1 "user gets convoluted message", what do you think?

xulman:
The fix dialog path, in its "dummy button", got option to adjust the dummy data to be just "around the spots" (aka axis aligned bounding box in spatial axes and temporal axis)... 
Screenshot_20220919_190545

you can find it in [tomancak mastodon repo, branch aabb-dummy](https://github.com/mastodon-sc/mastodon-tomancak/tree/aabb-dummy)

tinevez:
Hello @xulman 
I think this PR is ready to be merged. 
Can you briefly tell me of a procedure that would allow me to test it?

xulman:
- take the latest from https://github.com/xulman/mastodon/tree/enableOpeningOfDummyData (which is a fork of the official main mastodon/dev branch on which 4 commits to handle the dummy data are added)
- take the latest commit from [mastodon-tomacak repo](https://github.com/mastodon-sc/mastodon-tomancak), master branch
- compile & run
- take any mastodon project, copy it and cripple the path to the BDV data in the copy (e.g. introduce typo in the path in the project.xml or just move the data itself to another folder)
- try to open this project...

PR #195, opened 2022-09-15
ADD: MastodonDndLauncher to accept DnD .mastodon events

xulman:
HI, here's  a code that enables drag and drop of `.mastodon` files.
I somehow thought I was pushing it before but now I couldn't find it.. so here's #2...
(and it wasn't functional on current Mastodon Fiji distro anyway)
V.

### DIFF
--- a/src/main/java/org/mastodon/mamut/launcher/MastodonDndLauncher.java
+++ b/src/main/java/org/mastodon/mamut/launcher/MastodonDndLauncher.java
@@ -1,0 +1,68 @@
+package org.mastodon.mamut.launcher;
+
+import org.mastodon.mamut.MainWindow;
+import org.mastodon.mamut.WindowManager;
+import org.mastodon.mamut.project.MamutProject;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mastodon/mamut/launcher/MastodonDndLauncher.java
```diff
@@ -0,0 +1,68 @@
...
+		System.setProperty( "apple.laf.useScreenMenuBar", "true" );
+		final WindowManager windowManager = new WindowManager( getContext() );
+
+		final MainWindow mainWindow = new MainWindow( windowManager );
+		mainWindow.setVisible( true );
+		mainWindow.setDefaultCloseOperation( WindowConstants.EXIT_ON_CLOSE );
+
+		try {
+			final MamutProject project = new MamutProjectIO().load( projectPath );
+			windowManager.getProjectManager().open( project, true );
```
xulman, line 52: ....not sure if this whole code sequence is the correct way to start mastodon, please review @tinevez

PR #196, opened 2022-09-19
Idea: RecentProjectsPanel opens FileBrowser on dbl-click

xulman:
Sometime one has multiple `.mastodon` projects nearby in otherwise deeply nested common folder, one clicks all the way to open the first project.. later one wants to open the second project... and must click through again (or Cut&Paste the path, edit aside, open dialog, Paste-in the edited path, choose `.mastodon` file)

Or, one opens a file browser (by dbl clicking), then navigates to the wanted nearby project, then *Open another project* in Launcher which opens Dialog into which the path to the wanted nearby project can often be dnd'ed from the Browser...

**This is only a proposal.**

### DIFF
--- a/src/main/java/org/mastodon/mamut/launcher/RecentProjectsPanel.java
+++ b/src/main/java/org/mastodon/mamut/launcher/RecentProjectsPanel.java
@@ -4,6 +4,13 @@ import java.awt.Font;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
+import java.awt.Desktop;
+import java.awt.event.MouseEvent;
+import java.awt.event.MouseListener;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.function.Consumer;
 
@@ -82,6 +89,12 @@ public class RecentProjectsPanel extends JPanel
 
 		if ( recentProjects != null && !recentProjects.isempty() )
 		{
+			gbc.gridy++;
+			final JLabel lblTitleHint = new JLabel( "Double-click to open the containing folder." );
+			lblTitleHint.setFont( lblTitleHint.getFont().deriveFont( lblTitleHint.getFont().getStyle() | Font.ITALIC ) );
+			lblTitleHint.setHorizontalAlignment( SwingConstants.CENTER );
+			add( lblTitleHint, gbc );
+
 			for ( final String projectPath : recentProjects )
 			{
 				gbc.gridy++;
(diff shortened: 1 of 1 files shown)


xulman:
This is what i can look like:
Screenshot_20220919_174828

tinevez:
Does it work well when you have mastodon projects that are actually folders?

src/main/java/org/mastodon/mamut/launcher/RecentProjectsPanel.java
```diff
@@ -82,13 +89,20 @@ private void remakeGUI( final Consumer< String > projectOpener )
...
+			add( lblTitleHint, gbc );
+
 			for ( final String projectPath : recentProjects )
 			{
 				gbc.gridy++;
 				gbc.gridx = 0;
 				final JTextArea ta = new JTextArea( projectPath );
 				ta.setEditable( false );
 				ta.setLineWrap( true );
+				ta.addMouseListener( new MouseDblClickOpenPath( projectPath ) );
```
xulman, line 105: action no. 1: the mouse click on the text field is activated here, here it all starts

it creates a dedicated mouse listener for exactly this particular text field item -> see action no. 2

it passes further the string with the text as it is in the text field
the string can (theoretically) be

- a path to a regular file `.mastodon`
- a path to a regular folder

src/main/java/org/mastodon/mamut/launcher/RecentProjectsPanel.java
```diff
@@ -140,4 +154,44 @@ private void remakeGUI( final Consumer< String > projectOpener )
...
+
+	public static final class MouseDblClickOpenPath implements MouseListener {
+		final String Uri;
+		public MouseDblClickOpenPath(final String pathOnDblClick) {
+			Uri = Paths.get(pathOnDblClick).getParent().toUri().toString();
+		}
+
+		@Override
+		public void mouseClicked(MouseEvent mouseEvent) {
+			if (mouseEvent.getClickCount() == 2) openUrl( Uri );
```
xulman, line 167: action no. 3: here is the click processing, actually must be a dbl-click
it opens the folder at the memorized path (the parent folder of the dlg item content)

src/main/java/org/mastodon/mamut/launcher/RecentProjectsPanel.java
```diff
@@ -140,4 +154,44 @@ private void remakeGUI( final Consumer< String > projectOpener )
 		revalidate();
 		repaint();
 	}
+
+
+	public static final class MouseDblClickOpenPath implements MouseListener {
+		final String Uri;
+		public MouseDblClickOpenPath(final String pathOnDblClick) {
+			Uri = Paths.get(pathOnDblClick).getParent().toUri().toString();
```
xulman, line 162: action no. 2: here is the dedicated mouse listener created, it memorizes a path for its later acting

this path is a **parent** path to the given one... so it opens

- a folder which holds the `.mastodon` file
- a folder which holds the project folder.. (so not the project's content but a folder with the project's folder)

xulman (commented):
the proposed solution shall open a parent folder for a project whose path is printed in the text fields of the main open wizzard dialog

so it opens a folder that holds the `.mastodon` or a "unzipped" folder with the Mastodon project

xulman:
...and I have also tested that (only on my Linux)

PR #203, opened 2022-11-04
Add classes for starting Mastodon from the IDE for easy debugging

maarzt:
This PR adds three classes. These classes are in the test folder. They are therefor not part of the production code. They are instead meant for debugging Mastodon within the IDE (Eclipse, IntelliJ etc.)

These classes are:
1. **StartImageJ** Shows the ImageJ main window with Mastodon installed.
2. **StartMastodonLauncher** Shows ImageJ main window and the Mastodon launcher.
3. **StartMastodonOnProject** Shows a file open dialog that allows to select a mastodon project. Mastodon is started on the given project.

### DIFF
--- a/src/test/java/org/mastodon/StartImageJ.java
+++ b/src/test/java/org/mastodon/StartImageJ.java
@@ -1,0 +1,47 @@
+/*-
+ * #%L
+ * Mastodon
+ * %%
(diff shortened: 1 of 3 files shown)


maarzt:
To my opinion it's very good to have these classes in place. Whenever somebody new joins the Mastodon developing. They can simple execute one of these classes to debug Mastodon in the development environment.

PR #209, opened 2022-11-11
Unit test, to see if mastodon can deal with windows / linux saved file paths (Issue 205)

### LINKED ISSUE #205: Windows - Linux - Compatibility
# User Story

Mastodon Users would like to be able to exchange project files between different operating systems. More specifically, Windows, Linux, MacOS.

# Acceptance

1. Project file can be saved from Linux environment and then opened from Windows environment.
1. Project file can be saved from Windows environment and then opened from Linux environment.

# Tasks

- [x] Test, if this actually works already.
- ~~Save path of SpimDataFile on project.xml in Linux-style, i.e. using forward slashes ("/")~~
grafik
- ~~Ensure that Linux-style paths are read correctly from Windows environments.~~
- [x] Write Unit test that allows checking the behaviour from different OS machines.

stefanhahmann:
Resolves #205

### DIFF
--- a/src/test/java/org/mastodon/views/trackscheme/bdv/SharedBigDataViewerDataTest.java
+++ b/src/test/java/org/mastodon/views/trackscheme/bdv/SharedBigDataViewerDataTest.java
@@ -1,0 +1,33 @@
+package org.mastodon.views.trackscheme.bdv;
+
+import bdv.viewer.ViewerOptions;
+import mpicbg.spim.data.SpimDataException;
+import org.junit.Test;
+import org.mastodon.views.bdv.SharedBigDataViewerData;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+
+public class SharedBigDataViewerDataTest
+{
+	@Test
+	public void testFromSpimDataXmlFileWindowsStyle() throws SpimDataException, IOException
+	{
+		ViewerOptions viewerOptions = new ViewerOptions();
+		SharedBigDataViewerData sharedBigDataViewerData = SharedBigDataViewerData.fromSpimDataXmlFile( "./src/test/resources/org/mastodon/mamut/importer/trackmate/mamutproject/FakeTracksBDV.xml", viewerOptions, null );
+		assertEquals( 50, sharedBigDataViewerData.getNumTimepoints() );
+		assertEquals( 128, sharedBigDataViewerData.getSpimData().getSequenceDescription().getViewSetupsOrdered().get( 0 ).getSize().dimension( 0 ) );
+		assertEquals( 128, sharedBigDataViewerData.getSources().get( 0 ).getSpimSource().getSource( 0, 0 ).dimension( 0 ) );
+	}
+
+	@Test
+	public void testFromSpimDataXmlFileUnixAndMacOSStyle() throws SpimDataException, IOException
+	{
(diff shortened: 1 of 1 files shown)


stefanhahmann:
@maarzt may I ask you to run the 2 new tests on Linux (Ubuntu) machine.

maarzt:
Yes, it works. But I don't like that the test relies on exceptions to be thrown. I need to show you how to create these BDV XML + HDF5 datasets. We could put a tiny dataset into the test/resources folder.

maarzt:
@stefanhahmann here is a link on how to create the XML + HDF5 datasets

stefanhahmann:
> Yes, it works. But I don't like that the test relies on exceptions to be thrown. I need to show you how to create these BDV XML + HDF5 datasets. We could put a tiny dataset into the test/resources folder.

Good idea. Would be a better test then.

src/main/java/org/mastodon/views/bdv/SharedBigDataViewerData.java
```diff
@@ -370,6 +370,8 @@ public static SharedBigDataViewerData fromSpimDataXmlFile(
 			{
 				//trying to load the actual (not-dummy) data
 				spimData = new XmlIoSpimDataMinimal().load( spimDataXmlFilename );
+
+				System.out.println( "Successfully loaded spim data file: " + spimDataXmlFilename );
```
maarzt, line 374: This ```system.out.println(...)``` should be removed

src/test/java/org/mastodon/views/trackscheme/bdv/SharedBigDataViewerDataTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class SharedBigDataViewerDataTest
+{
+	@Test
+	public void testFromSpimDataXmlFileWindowsStyle() throws SpimDataException, IOException
+	{
+		ViewerOptions viewerOptions = new ViewerOptions();
+		SharedBigDataViewerData sharedBigDataViewerData = SharedBigDataViewerData.fromSpimDataXmlFile( "./src/test/resources/org/mastodon/views/trackscheme/bdv/dataset_hdf5.xml", viewerOptions, null );
```
maarzt, line 18: In noticed, that there is already a BDV XML+H5 in the resources folder which we could use:
```src/test/resources/org/mastodon/mamut/importer/trackmate/mamutproject/FakeTracksBDV.xml```

maarzt (commented):
I suggested two changes

src/main/java/org/mastodon/views/bdv/SharedBigDataViewerData.java
stefanhahmann: Removed.

src/test/java/org/mastodon/views/trackscheme/bdv/SharedBigDataViewerDataTest.java
stefanhahmann: Thanks for the hint. Made changes to use these resources.

maarzt:
Sorry this is now fixed in #210 and there is also a new unit test.

I close this PR in favor of #210

PR #210, opened 2022-11-21
Improve dummy data when opening mastodon project with missing images (Issue 206)

### LINKED ISSUE #206: Improve dummy data when loading mastodon project without image data
## User Story

Users often transfer only *.mastodon files without the image data. This is done for practical reasons, as the image files might be very large and are therefore hard to share. A user would like to be able to open *.mastodon files without the actual image data available. This is currently quite complicated and should be improved.

### Current complicated user interaction:
1. User opens a *.mastodon file without image data present.
2. A dialog pops up to tell the user that dummy data is loaded instead:
image
3. The dummy data always has only a fixed number of timepoints. This means the TrackScheme only displays a part of the data and is therefore not usable:
image
4. The BigDataViewer (BDV) is in similar way barely functional:
image
5. In order to fix this problem the user needs to follow the instructions of dialog in step (2): Go to "File -> Fix Image Path..." this dialog is part of another git project "mastodon-tomancak", and might sometime not be installed. It allows users to semi-automatically enter pixel sizes and image sizes. Finally the "image path" in the project file is replaced by a dummy string.
6. Mastodon needs to be closed an the project needs to be reopened.

### Ideal user interaction:
1. User opens a *.mastodon file without image data present.
2. A dialog pops up to tell the user that dummy data is loaded.
3. Mastodon opens the project with dummy data, that has appropriate size and time points, to use TrackScheme and BDV.
4. _(Ideally the dummy data also has the same pixel size & image size & transformation as the original data, so that plugins relying on this information work as expected.)_

## Acceptance criteria

1. Mastodon Plugin saves a copy of the SpimDataFile directly into the mastodon project.
1. When opening a Mastodon project, it is checked, if the actual/original SpimDataFile is available. If it is not available the copy of the SpimDataFile within the Mastodon project is read.
1. In that case the user is informed, that no image data is available, but that lineage data can still be analysed. Relevant parameters, such as pixel size, image size are read from the copy of the SpimDataFile
1. Generate appropriate Dummy data, i.e. black background with appropriate size and appropriate number of time segments

## Tasks
- [x] Save SpimDataFile to the Mastodon project, when a Mastodon project is saved. Suggested file name: "dataset.xml.backup"
- [x] When opening a Mastodon project and no SpimDataFile is available, try to read "dataset.xml.backup" from Mastodon project file
- [x] Show a dialog informing the user: "No image available. Relevant meta information was read from Mastodon project file. You will be able to work with the lineage data. Do you want to procced?" "Yes/No"
- [x] In the "Yes" case, the image sizes, pixel sizes, and image transform parameters are read from the "dataset.xml.backup" file.  Instead of the intensity image, a black dummy image is loaded that has the same pixel transformation as the original image. _This results in reading the dataset.xml without the *.tif or *.h5 files present. To do so, it is probably necessary to replace the standard image loaders with dummy image loaders, this might require changes to the BDV libraries._
- [x] Path of SpimDataFile remains the same the same and is not updated.

## Additional Information

- Path if SpimDataFile stored in project file
grafik

- Example content of a SpimDataFile

grafik

- Documention of Discussion Matthias-Stefan:

stefanhahmann:
Resolves #206


stefanhahmann:
@maarzt this pull request is in draft mode for now. please test it.

maarzt:
Hello Stefan,

It works exactly as we discussed it, that's great. I didn't had a look into the code for now. But I noticed that the message dialog that is shown to the user is a little bit confusing:
image

Maybe a dialog like this would be better:

- **Title:** Mastodon - Image data not found.
- **Icon:** Warning Message
- **Text:**
  Mastodon could not find the images associated with this project.
This file was not found:
/path/to/file/not/found/dataset.xml

  It is still possible to open the project. You won't be able to see the image data, but you can still view and modify the tracking data.

  Would you like to open the project without image data?
- **Buttons:** [Open Project] [Cancel]

stefanhahmann:
> Hello Stefan,
> 
> It works exactly as we discussed it, that's great. I didn't had a look into the code for now. But I noticed that the message dialog that is shown to the user is a little bit confusing: image
> 
> Maybe a dialog like this would be better:
> 
>     * **Title:** Mastodon - Image data not found.
> 
>     * **Icon:** Warning Message
> 
>     * **Text:**
>       Mastodon could not find the images associated with this project.
>       This file was not found:
>       /path/to/file/not/found/dataset.xml
>       It is still possible to open the project. You won't be able to see the image data, but you can still view and modify the tracking data.
>       Would you like to open the project without image data?
> 
>     * **Buttons:** [Open Project] [Cancel]

Good idea. Applied suggested changes in a further commit.

stefanhahmann:
Visualisation of new behaviour:

* issue-216-1
* issue-216-2

src/main/java/org/mastodon/mamut/project/MamutProject.java
```diff
@@ -204,6 +206,8 @@ public interface ProjectWriter extends Closeable
 		OutputStream getFeatureOutputStream( String featureKey ) throws IOException;
 
 		OutputStream getGuiOutputStream() throws IOException;
+
+		OutputStream getDatasetXmlFileOutputStream() throws IOException;
```
maarzt, line 212: Please rename to `getDatasetXmlOutputStream` to be more consistent with the other methods

src/main/java/org/mastodon/mamut/ProjectManager.java
```diff
@@ -857,14 +863,16 @@ private void saveGUI( final ProjectWriter writer ) throws FileNotFoundException,
 		final Element guiRoot = new Element( GUI_TAG );
 		guiRoot.setAttribute( MAMUTPROJECT_VERSION_ATTRIBUTE_NAME, MAMUTPROJECT_VERSION_ATTRIBUTE_CURRENT );
 		final Element windows = new Element( WINDOWS_TAG );
-		windowManager.forEachView( ( view ) -> windows.addContent(
-				MamutViewStateSerialization.toXml( view ) ) );
-		windowManager.forEachBranchView( ( view ) -> windows.addContent(
-				MamutViewStateSerialization.toXml( view ) ) );
```
maarzt, line 863: It's hard to spot what actually changed, since there are so many lines just reformatted could you undo the formatting please!

src/main/java/org/mastodon/mamut/project/MamutProject.java
```diff
@@ -246,8 +248,7 @@ public Collection< String > getFeatureKeys()
 
 			final List< String > featureKeys = Arrays.stream( featureFolder.listFiles( ( dir, name ) -> name.toLowerCase().endsWith( ".raw" ) ) )
 				.map( f -> f.getName() )
-				.map( s -> s.replace( ".raw", "" ) )
-				.collect( Collectors.toList() );
+				.map( s -> s.replace( ".raw", "" ) ).collect( Collectors.toList() );
```
maarzt, line 251: Please undo the code reformatting

src/main/java/org/mastodon/mamut/launcher/MastodonLauncher.java
```diff
@@ -627,7 +627,13 @@ private void loadMastodonProject( final String projectPath )
 					try
 					{
 						final MamutProject project = new MamutProjectIO().load( file.getAbsolutePath() );
-						windowManager.getProjectManager().open( project, true );
+						boolean projectSuccessfullyOpened = windowManager.getProjectManager().open( project, true );
```
maarzt, line 630: I don't like this boolean return value. My first idea would be to throw a CancellationException instead and catch it in appropriate places. I'm open to discuss this.

I see two use-cases for opening a Mastodon project: 1. with UI & user interaction 2. headless in a script.
And the code less and less supports the second usecase... (I don't know a good solution yet)

src/main/java/org/mastodon/views/bdv/SharedBigDataViewerData.java
```diff
@@ -354,21 +355,57 @@ private static final String errorMessageUnknownHost( final String xmlFilename, f
...
 		AbstractSpimData< ? > spimData = DummySpimData.tryCreate( spimDataXmlFilename );
+		// SpimData is still null, if loading dummy data was not specified in the spim data file. This should be the "normal" situation
 		if ( spimData == null )
 		{
 			try
 			{
-				//trying to load the actual (not-dummy) data
+				File spimDataXmlFile = new File( spimDataXmlFilename );
+				if ( !spimDataXmlFile.exists() )
+				{
```
maarzt, line 374: please refactor into a smaller method

src/main/java/org/mastodon/mamut/ProjectManager.java
```diff
@@ -545,6 +547,20 @@ public synchronized void saveProject( final File projectRoot ) throws IOExceptio
...
+		{
+			backupSpimDataFile = File.createTempFile( "Mastodon-SpimBackup-", ".tmp" );
+			backupSpimDataFile.deleteOnExit();
+			Files.copy( reader.getDatasetXmlFileInputStream(), backupSpimDataFile.toPath(), StandardCopyOption.REPLACE_EXISTING );
+		}
+		catch ( IOException ignored )
+		{
+			backupSpimDataFile = null;
+		}
+
```
maarzt, line 562: This should be moved into the saveDataFileBackupm method.

maarzt (changes requested):
I very much like how well this works. I think it's a big improvement in terms of user experience.

There are still some things that I would like to improve. The code for opening a mastodon project is already quite complex, and would like to avoid increasing the complexity. Could you do the requested changes. I would afterwards continue, try to refactor some stuff.

tinevez:
Hello @stefanhahmann @maarzt 

I am only catching up now. I think that this PR may be duplicating work already done, by @xulman 

The beta-26 can deal with missing BDV data but improperly. For instance the number of time-points is not detected correctly and the dummy data implemented with Vlado seems not to work in full (you need to resave the file for the dummy to work). 
I have a gigantic dataset that starts at frame 400, but I cannot simply display the tracks nor TrackScheme if I do not have the BDV.

Maybe you want to test Vlado solution? 

In Fiji, open Mastodon making sure we use the beta-26. We need the Fiji version, because it has the mastodon-tomancak artifact that contains the file path editor of Vlado. 

Open a Mastodon file that points to a missing XML BDV file. Mastodon will warn that it cannot find the associated bdv file.
Go to the _File > Fix image path_ menu item. There is a "I want dummy data instead", then "from spots". This will fix the issue, but only if you save and reload. 

Please tell us what you think.

maarzt:
Hello @tinevez,

We are well aware of the work done so far be @xulman. This PR continues where this previous work stopped.

The goal is to simplify the user interaction. It is no longer required to re-save the Mastodon project and users don't need to enter information like pixel-size etc.

I would also like to move the code in mastodon-tomancak over to the main repository. As it would make the mastodon git repository more self contained.

The PR might fix the problems with the dataset that starts at frame 400. Could you maybe give this a try?

Bests, Matthias

tinevez:
Yes ok! 
When the PR is ready we will ask Vlado if he is ok with transitioning his code with yours to the core, if that's ok.

src/main/java/org/mastodon/mamut/project/MamutProject.java
stefanhahmann: Done.

stefanhahmann:
> @stefanhahmann Please write a few sentences, how opening the spim data XML works, if the HDF5 data is missing. I only understand parts of this. Like that the ImgLoader part has the removed from the XML, and there are some null values that need to be replaced later on. I think it would be useful having this documented in the issue or somewhere.

As discussed: No changes on the spim data XML are necessary (initially, I thought, they would be necessary, but in fact it turned out that they weren't). The process works now more or less as described in the acceptance criteria of https://github.com/mastodon-sc/mastodon/issues/206

stefanhahmann (commented):
I made some commits to consider your review comments.
Could you please review again?

maarzt:
Hi @tinevez,

Let me summarize this PR for you, such that you have an easier time reviewing it. The goal of the PR is to streamline opening a Mastodon project, when the image data is missing. The changes will make it a lot easier to open a Mastodon project that one just got send via email, where image data is missing as usual...

The current version of Mastodon, can be started if the image data is missing. But it only starts with partly functional dummy data. In this state, only the first ten time-points of the trackscheme are visible and a new BDV window is zoomed suboptimally, since the image size incorrect. The user needs to click "File > Fix Image Path", resave and restart Mastodon in order to conveniently inspect the data.

This PR improves this situation very much. The main idea is to save a copy of the spim data XML within the Mastodon project. This backup file is called "dataset.xml.backup". If now a Mastodon project is opened and the primary image data is missing, Mastodon can read the "dataset.xml.backup" file to initialize dummy image data with the correct number of time points, image sizes and even affine image transformations. The TrackScheme and BDV windows work as expected with the now correct dummy image data.

The PR also improves the dialog that tells the user about the problem loading the image data. The dialog now tells the user what went wrong. And there is also a convenient "cancel" button.

Here is an overview of the changes made in the code:
* New method `DummySpimData.fromSpimDataXml()` allows to create dummy data with correct metadata from a spim data XML. This works as follows:
  * Create a temporary copy of the spim data XML, with the `<ImageLoader>` entry removed
  * Open this modified XML. The result is a `SpimData` object with `imgLoader` not set.
  * Finally `imgLoader` is set to use the `DummyImgLoader`.
  * Et voila: we got a `SpimData` object with all metadata set correctly, buy no actual image pixel data required or loaded.
* ProjectManager.openImageData(...) and SharedBigDataViewerData.fromSpimDataXml(...) have been refactored.
* The dialog telling the user about problems loading image data is now in the new method ProjectManager.openWithDialog(...).
  This still allows to open a project without dialog using ProjectManager.open(...), which is useful for batch processing.
* MamutProjectIO is now responsibly for correctly loading Windows file paths with backslashes on Linux
* Newly added tests:
  * ProjectManagerTest: Test if the dataset.xml.backup is saved correctly.
  * StartMastodonWithMissingImageDemo: Demonstrate how Mastodon starts if image data is missing.
  * MamutProjectIOTest: Make sure Mastodon can deal with Windows and Linux file paths.
  * DummySpimDataTest & DummyImgLoaderTest

Bests,
Matthias

maarzt (approved):

tinevez (approved):

tinevez:
Hello all. High quality code! 
Indeed, this new version does handling missing data really well and I would say, the proper way. The idea of creating a temp file is very good and very efficient. I tested the feature with image data hosted remotely and it works as expected. 
Next step would probably be merging the tool to fix the image path from Vlado into the core.

PR #215, opened 2022-12-02
Color TrackScheme Hierarchy according to the time point (Issue 213)

### LINKED ISSUE #213: Color TrackScheme Hierarchy according to the time point
# User Story

As a Mastodon user, I may use the TrackScheme Hierarchy view to get a generalised overview of the lineage tree. This view does not contain any individual spots and is not scaled according to the time model making it easier for the user to investigate the lineage tree structure as a whole.
It may still be useful for the user, if the lineage tree could be rendered in a way that highlights branches that already exist at a certain timepoint while navigating through the available time steps in other views.

# Acceptance criteria

1. The Branches in the TrackScheme Hierarchy that do not exist at a time point selected via BigDataViewer or TrackScheme are drawn in a faded color, e.g.: 
grafik
1. This feature is only available if the TrackScheme Hierarchy View is coupled to other views: 
grafik
1. Original branch color is used in this case: TimePointModel.TimePoint >= BranchSpot.FirstTimePoint
1. A mixture of background color and branch color is used in all other cases

# Tasks
- [x] Add a reference to the real time point model to the class MamutBranchViewTrackSchemeHierarchy in addition to the already existing reference to the dummy time point model
- [x] Use the reference to the real time point model (if exists) to color branches according to the current settings in the real time point model
- [x] Add option to toggle this behaviour on/off for TrackSchemeHierarchy View via Preferences

# Additional information:

* One TrackSchemePanel for views TrackScheme, TrackScheme Branch and TrackScheme Hierarchy
* The TrackSchemePanel has multiple Renderers: 
  * The Renderer for TrackScheme Hierarchy View is in the class PaintHierarchicalGraph
  * This class extends PaintGraph
  * PaintGraph is a Graph Renderer
* There is also the ModelGraph - this is the representation of the acutal lineage graph structure
* There is the TrackSchemeGraph. It is a Listener on the ModelGraph and defines it visualisation. The TrackSchemeGraph copies the ModelGraph. There is one TrackSchemeGraph per Window, thus enabling different TrackScheme colorings in different windows.
* There is the ModelBranchGraph. It is a Listener on the ModelGraph that contains the simplified structure of the ModelGraph that is used in the TrackSchemeHierarchy View
* There is a TimePointModel that contains the different timepoints of the Mastodon project
* The MamutBranchViewTrackSchemeHierarchy has a Dummy TimePointModel to enable its time independent visualisation
* The MamutBranchViewTrackSchemeHierarchy additionally needs some reference to the real time point model
* Each Spot is aware of its timepoint
* The MamutBranchViewTrackSchemeHierarchy extends MamutBranchViewTrackSchemeBranch (these are the windows)

stefanhahmann:
Closes #213


stefanhahmann:
Visualisation of current implementation:

issue213

grafik

stefanhahmann:
Feedback from users:

* both daughter cells should get black at the same time
* could we have the same feature also for branch view? → actually not so much needed, since there is already the horizontal line
* would be nice to have the sum of the spots of the current timepoint somewhere visible in the trackscheme (e.g. in the bottom bar of the trackscheme view)

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
```diff
@@ -197,17 +199,18 @@ public TrackSchemePanel(
 			final TrackSchemeGraph< ?, ? > graph,
 			final HighlightModel< TrackSchemeVertex, TrackSchemeEdge > highlight,
 			final FocusModel< TrackSchemeVertex, TrackSchemeEdge > focus,
-			final TimepointModel timepoint,
+			final TimepointModel navigationTimepointModel, final TimepointModel fadingTimepointModel,
```
maarzt, line 202: Please add a new line after the comma, to keep the rule of one parameter per line.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemeFrame.java
```diff
@@ -64,7 +64,7 @@ public TrackSchemeFrame(
 			final TrackSchemeGraph< ?, ? > graph,
 			final HighlightModel< TrackSchemeVertex, TrackSchemeEdge > highlight,
 			final FocusModel< TrackSchemeVertex, TrackSchemeEdge > focus,
-			final TimepointModel timepoint,
+			final TimepointModel navigationTimepointModel, final TimepointModel fadingTimepointModel,
```
maarzt, line 67: Please add a new line after the comma, to keep the rule of one parameter per line.

src/main/java/org/mastodon/views/trackscheme/LineageTreeLayoutImp.java
```diff
@@ -150,10 +156,11 @@ public class LineageTreeLayoutImp implements LineageTreeLayout
 	public LineageTreeLayoutImp(
 			final RootsModel< TrackSchemeVertex > rootsModel,
 			final TrackSchemeGraph< ?, ? > graph,
-			final SelectionModel< TrackSchemeVertex, TrackSchemeEdge > selection )
+			final SelectionModel< TrackSchemeVertex, TrackSchemeEdge > selection, final @Nullable TimepointModel timepointModel )
```
maarzt, line 159: Please add a new line after the comma, to keep the rule of one parameter per line.

maarzt (commented):
Only minor comments so far.

There is still a bug: Spot and edges are not faded simultaneously, that's strange.

And the ratio for mixing colors does not work on my screen, I don't see the faded spots & edges. We may need to allow the user to adjust this ratio.

I'm not yet fully done with the review. I will have a deeper look, why the ColorGenerator can not be used and if there is less confusing way than having two TimepointModel... let's see.

src/main/java/org/mastodon/views/trackscheme/LineageTreeLayoutImp.java
```diff
@@ -431,43 +442,45 @@ public void cropAndScale( final ScreenTransform transform, final ScreenEntities
...
+		final String label = trackSchemeVertex.getLabel();
+		final boolean selected = selection.isSelected( trackSchemeVertex );
+		final boolean ghost = trackSchemeVertex.isGhost();
 		// TODO move setYStart into init
-		screenVertexPool.create( sv ).init( id, label, x, y, selected, ghost, colorGenerator.color( v1 ) )
+		screenVertexPool.create( screenVertex ).init( id, label, x, y, selected, ghost, colorGenerator.color( trackSchemeVertex ) )
 				.setYStart( firstY );
-		screenVertices.add( sv );
+		if ( timepointModel != null )
+			screenVertex.setFaded( trackSchemeVertex.isBeforeTimepoint( timepointModel.getTimepoint() ) );
```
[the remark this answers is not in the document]
maarzt: Maybe the ColorGenerator can be used instead. But I think I remember that you already tried, and we somehow where not happy with the visualization.

maarzt (commented):
The code changes made in this PR should be improved in terms of software architecture. Here are some potential ideas: 

1. The property of `ScreenVertices` to be faded is somehow similar to color and selected properties. I think there should be a `FadedModel` like there is a `SelectionModel` of a `GraphColorGenerator`. We would also need a `FadedAdapter` like there is a `SelectionAdapter`. The `isBeforeTimepoint` and `isAfterTimepoint` currently fulfill the task of this `FadedModel`, but this solution appears wrong to me, because now suddenly the logic that vertices with isBeforeTimepoint(time) are faded is expressed in the `LineageTreeLayoutImp` were it doesn't belong.
3. (Idea) It is maybe nice if instead of a "faded" flag, we have a "opacity" value 0-255 in the `ScreenVertex` and `ScreenEdge`. (The `FadedModel` I talked above should then be called `OpacityModel`). This would open many possible uses for the `OpacityModel`  in the future.
4. Maybe we can also simplify this all together, and use the alpha value in the `ScreenVertex.color` property as `opacity`. This would make the `GraphColorGenerator` powerful enough, such that we no longer need any `faded`, `opacity` property and 'OpacityModel` etc.

And a feature request in terms of usability:

It must be possible for a user to select the ratio used for fading vertices. This is necessary to support more TrackSchemeStyle color schemes.

src/main/java/org/mastodon/util/ColorUtils.java
```diff
@@ -0,0 +1,69 @@
...
+		int r2 = ( ( i2 & 0xff0000 ) >> 16 );
+		int g2 = ( ( i2 & 0xff00 ) >> 8 );
+		int b2 = ( i2 & 0xff );
+
+		int a = ( int ) ( ( a1 * iRatio ) + ( a2 * ratio ) );
+		int r = ( int ) ( ( r1 * iRatio ) + ( r2 * ratio ) );
+		int g = ( int ) ( ( g1 * iRatio ) + ( g2 * ratio ) );
+		int b = ( int ) ( ( b1 * iRatio ) + ( b2 * ratio ) );
+
+		return new Color( a << 24 | r << 16 | g << 8 | b );
```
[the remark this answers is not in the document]
maarzt: ok, I made a commit, that implements the requested changes.

src/main/java/org/mastodon/views/trackscheme/LineageTreeLayoutImp.java
maarzt: done

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
stefanhahmann: Done (in the new implementation with the FadedModel).

src/main/java/org/mastodon/views/trackscheme/wrap/ModelGraphProperties.java
```diff
@@ -56,6 +56,10 @@
 {
 	public int getTimepoint( V v );
 
+	boolean isBeforeTimepoint( V v, int timepoint );
+
+	boolean isAfterTimepoint( V v, int timepoint );
```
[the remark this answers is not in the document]
stefanhahmann: This implemented in #231 now.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemeFrame.java
maarzt: Awesome. I find this so much better now. That we don't need two timepoint models here.

PR #231, opened 2023-03-27
Color TrackScheme Hierarchy according to the time point (Issue 213)

### LINKED ISSUE #213: Color TrackScheme Hierarchy according to the time point

stefanhahmann:
Closes #213


src/main/java/org/mastodon/views/trackscheme/LineageTreeLayoutImp.java
```diff
@@ -87,6 +90,11 @@ public class LineageTreeLayoutImp implements LineageTreeLayout
 
 	protected final SelectionModel< TrackSchemeVertex, TrackSchemeEdge > selection;
 
+	protected final GraphColorGenerator< TrackSchemeVertex, TrackSchemeEdge > colorGenerator;
+
+	@Nonnull
```
maarzt, line 95: Please remove the @Nonnulll annotation for the sake of consistency. All other fields are expected to be not null too.

src/main/java/org/mastodon/views/trackscheme/display/PaintGraph.java
```diff
@@ -497,37 +512,23 @@ else if ( specifiedColor == 0 )
...
 			return isGhost
 					? TrackSchemeStyle.mixGhostColor( color, style.getBackgroundColor() )
 					: color;
 		}
 	}
 
+	protected boolean isFadingFutureTimepoints()
+	{
+		return false;
+	}
```
maarzt, line 530: This method here is part of a hack. You inject the code that is actually needed for the PaintHierarchyGraph into this class but you dissable it by using the `isFadingGutureTimepoints()` method. Now there is a simple solution for this ScreenVertex.isFaded should always be false.

src/main/java/org/mastodon/views/trackscheme/display/style/TrackSchemeStyle.java
```diff
@@ -372,6 +374,11 @@ public boolean isHierarchyGraphCurvedLines()
 		return hierarchyCurvedLines;
 	}
 
+	public boolean isFadeFutureTimepoints()
+	{
+		return fadeFutureTimepoints;
+	}
+
```
maarzt, line 381: Having this setting in the TrackSchemeStyle feels wrong to me. It is too much hidden from the user. And it also feels like a misuse of the TrackSchemeStyle. I think now with the FadedModel it is possible to have a icon/button in the MamutBranchViewTrackSchemeHierarchy window, that enables or disables the FadedModel accordingly.

src/main/java/org/mastodon/mamut/MamutBranchViewTrackScheme.java
```diff
@@ -152,7 +158,8 @@ protected MamutBranchViewTrackScheme(
 				viewGraph,
 				highlightModel,
 				navigateFocusModel,
-				( timepointModel == null ) ? this.timepointModel : timepointModel,
```
maarzt, line 155: Removing this line will actually reintroduce a bug. Please don't remove that line. You may read the commit message that introduced this line to see what I'm talking about.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
```diff
@@ -177,7 +178,7 @@ public class TrackSchemePanel extends JPanel implements
 	 */
 	private boolean ignoreScrollBarChanges;
 
-	private final TimepointModel timepoint;
+	private final TimepointModel navigationTimepointModel;
```
maarzt, line 181: Why do you call this timepointModel navigationTimepointModel? Please undo the renaming if there's no reasoning behind it.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
```diff
@@ -197,17 +198,18 @@ public TrackSchemePanel(
...
 			final HighlightModel< TrackSchemeVertex, TrackSchemeEdge > highlight,
 			final FocusModel< TrackSchemeVertex, TrackSchemeEdge > focus,
-			final TimepointModel timepoint,
+			final TimepointModel navigationTimepointModel,
+			final FadedModel< TrackSchemeVertex, TrackSchemeEdge > fadedModel,
 			final SelectionModel< TrackSchemeVertex, TrackSchemeEdge > selection,
 			final RootsModel< TrackSchemeVertex > rootsModel,
 			final NavigationHandler< TrackSchemeVertex, TrackSchemeEdge > navigation,
-			final TrackSchemeOptions optional )
+			final TrackSchemeOptions trackSchemeOptions )
```
maarzt, line 206: "optional" is not a good name. I agree. But the class is TrackSchemePanel. Please name the variable simply "options".

src/main/java/org/mastodon/mamut/MamutBranchViewTrackScheme.java
```diff
@@ -98,6 +102,9 @@
 
 	private final ColoringModel coloringModel;
 
+	@Nullable
+	private JLabel timepointAndNumberOfSpotsLabel;
```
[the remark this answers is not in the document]
stefanhahmann: I have removed it, since this is unneeded. Apologies for this mistake. Will take more care in the future.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
stefanhahmann: There is already a local variable in this method called "options". I followed your advise to rename the name of the variable in the method body to options, but I had to rename the existing variable "options" to "optionValues" to make this possible.

src/main/java/org/mastodon/mamut/MamutBranchViewTrackScheme.java
stefanhahmann: Thanks for the hint. I added this line again together with a comment, explaining, why it is needed.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
stefanhahmann: There was a reasoning behind it during development. But that reason now does not exist anymore. Thus I undid the renaming.

src/main/java/org/mastodon/views/trackscheme/LineageTreeLayoutImp.java
```diff
@@ -150,10 +158,14 @@ public class LineageTreeLayoutImp implements LineageTreeLayout
 	public LineageTreeLayoutImp(
 			final RootsModel< TrackSchemeVertex > rootsModel,
 			final TrackSchemeGraph< ?, ? > graph,
-			final SelectionModel< TrackSchemeVertex, TrackSchemeEdge > selection )
+			final SelectionModel< TrackSchemeVertex, TrackSchemeEdge > selection,
+			final GraphColorGenerator< TrackSchemeVertex, TrackSchemeEdge > colorGenerator,
+			final @Nonnull FadedModel< TrackSchemeVertex, TrackSchemeEdge > fadedModel )
```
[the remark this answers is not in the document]
stefanhahmann: Removed it.

src/main/java/org/mastodon/views/trackscheme/display/PaintGraph.java
stefanhahmann: I removed the method.

src/main/java/org/mastodon/mamut/MamutBranchViewTrackScheme.java
stefanhahmann: I failed to reproduce the bug that is described here (https://github.com/mastodon-sc/mastodon/pull/201) when this line is removed. However, if change the line as you suggest, this introduces a bug into the fading.
Perhaps, you could show me how to reproduce the bug if that line is removed?

stefanhahmann: After checking again, I could added this line again.
There was another error, which I solved by adding a second timepoint model. Now, there are two timepoint models, one that is for highlighting the current row of spots (which is overridden in trackschemehierarchy) and one that is used for computing the fading.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemeFrame.java
```diff
@@ -79,7 +82,9 @@ public TrackSchemeFrame(
 				graph,
 				highlight,
 				focus,
-				timepoint,
+				timepointModel,
+				fadingTimepointModel,
+				fadedModel,
```
maarzt, line 87: The fadingTimepointModel isn't required here. Please add listeners to the fadedModel instead.

maarzt (changes requested):

src/main/java/org/mastodon/app/ui/TimepointAndNumberOfSpotsLabel.java
```diff
@@ -0,0 +1,45 @@
...
+ *  A label that displays (and updates according to the timepoint model) the current timepoint and the number of spots in the
+ *  current timepoint.
+ *  
+ * @author Stefan Hahmann 
+ */
+public class TimepointAndNumberOfSpotsLabel extends JLabel
+{
+	private final transient TimepointModel timepointModel;
+
+	private final transient Model model;
```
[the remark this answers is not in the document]
stefanhahmann: I agree. I removed it.

src/main/java/org/mastodon/app/ui/TimepointAndNumberOfSpotsLabel.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.swing.JLabel;
+import javax.swing.SwingUtilities;
+
+/**
+ *  A label that displays (and updates according to the timepoint model) the current timepoint and the number of spots in the
+ *  current timepoint.
+ *  
+ * @author Stefan Hahmann 
+ */
+public class TimepointAndNumberOfSpotsLabel extends JLabel
```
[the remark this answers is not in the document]
stefanhahmann: I changed it.
For further reference: why do you think a Panel having a Label is better than just a Label?

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemePanel.java
```diff
@@ -449,12 +454,13 @@ else if ( flags.entitiesAttributesChanged )
...
 	public void timepointChanged()
 	{
-		final int t = timepoint.getTimepoint();
+		final int t = timepointModel.getTimepoint();
 		if ( graphOverlay.getCurrentTimepoint() != t )
 		{
 			graphOverlay.setCurrentTimepoint( t );
 			display.repaint();
 		}
+		entitiesAttributesChanged();
```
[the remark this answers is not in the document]
stefanhahmann: Thanks for this advice. I was also unsure about this line, since I did not fully understand all the different flags that currently exist.
I added fadingChanged and the corresponding flad.

src/main/java/org/mastodon/views/trackscheme/display/TrackSchemeFrame.java
stefanhahmann: Thanks for the advice. I had this idea as well but did not find a solution before.
Now, I tried again. It is solved now via listeners() and the fadingTimepointModel is not needed here anymore. However, I am not sure, if my implementation meets your intention. Please check.

stefanhahmann: After this discussing this again I introduced a FadingListener that is implemented by the TrackSchemePanel.

src/main/java/org/mastodon/app/ui/TimepointAndNumberOfSpotsLabel.java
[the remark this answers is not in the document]
maarzt: It's the standard pattern to extend JPanel. Using JPanel is also a lot more flexible. Why would you want to extend JLabel?

src/main/java/org/mastodon/mamut/MamutBranchViewTrackScheme.java
maarzt: And now we finally don't even need the second timepoint model.

maarzt (approved):
I removed some unnecessary code, and moved a few other lines of code.
But overall the PR is ok now :relieved:

(I'm still worried about how many classes we needed to change to implement this one feature.)

maarzt:
@tinevez I'm ok with this being merged now. But the changes are very complex. So please have a look first.

src/main/java/org/mastodon/app/ui/TimepointAndNumberOfSpotsLabel.java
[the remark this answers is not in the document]
stefanhahmann: Keeping it simple.

My thought was to not use an extra JPanel where only a JLabel is actually needed. The potential extra flexibility of the JPanel is not needed the way this component is currently used. The JPanel in my view only adds an extra, yet unrequired, layer.

maarzt (changes requested):
The PR somehow introduces a bug when rendering the the TrackScheme Branch.
The bug is triggered by resorting the TrackScheme using one of the plugins in mastodon-tomancak. For example "Plugins > Trees Management > Flip Descendants".

GIF 05 05 2023 14-45-05

This need to be clarified before merging.

stefanhahmann:
> Ready for review?

Yes.
@maarzt reviewed already extensively, but since the changes are massive, a second review would be good.

maarzt (changes requested):

src/test/java/org/mastodon/util/ColorUtilsTest.java
```diff
@@ -29,10 +30,31 @@ public void testGetMixedColor()
...
+		assertEquals( 127, resultWithAlpha.getAlpha() );
+
+		// generate color from result without valid alpha bits
+		Color resultWithoutAlpha = new Color( ColorUtils.scaleAlpha( input.getRGB(), ratio ) );
+		assertEquals( 0, resultWithoutAlpha.getRed() );
+		assertEquals( 0, resultWithoutAlpha.getGreen() );
+		assertEquals( 0, resultWithoutAlpha.getBlue() );
+
+		// alpha bits are not set -> alpha is still (but wrongly) 255
+		assertEquals( 255, resultWithoutAlpha.getAlpha() );
```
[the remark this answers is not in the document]
maarzt: I removed the lines myself, and found another similar bug in `ColorUtils.mixColors(...)`.

Now there's a test to covers the alpha channel mixing in `ColorUtils.mixColors(...)`

maarzt (approved):

src/test/java/org/mastodon/util/ColorUtilsTest.java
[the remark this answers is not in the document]
stefanhahmann: I was kind of anticipating that you judge this code as being misplaced here. For this reason, I had put it in a separate commit ([ColorUtils.mixColors](https://github.com/mastodon-sc/mastodon/pull/231/commits/119c095d02459ca6b86c2e2cfba76fe237a03e11)), so that this commit could just be dropped.

However: I would like to keep this code somewhere, as it contributed very much to solving the bug related to the scaleAlpha() method. It also demonstrates how the result of scaleAlpha should be used, but I confess the unit test, may not be the best place for this. Maybe an improved javadoc would also already help.

stefanhahmann:
> Sorry I am late on this. Will work on it

@tinevez may we kindly remind you on this review? we would love to have this finished/merged.

tinevez:
I made a PR for your PR with minor changes, mainly sizing the opportunity to fix the look of the settings panel. 
Take them if you feel like it.
To me otherwise it is good to be merged to dev.

tinevez (approved):

stefanhahmann:
Hi @tinevez,

I merged your PR https://github.com/stefanhahmann/mastodon/pull/17.

However after merging the toolbar of the trackscheme windows looked like this on my (windows) machine:

grafik

Thus, I increased the size of the Group Locks Panel again: https://github.com/mastodon-sc/mastodon/pull/231/commits/316248399effe5b0d911e0022f2613f464874214

Also, I think, it is better to avoid abbreviations in user interfaces. Thus I made these changes: 

* https://github.com/mastodon-sc/mastodon/pull/231/commits/0deefa744d995f5ff45fdaaa515261cfa005e3f7
* https://github.com/mastodon-sc/mastodon/pull/231/commits/b54161374360039a49f51786da9a6e5ce4f600e5

which make the toolbar of trackscheme window look like this on my windows:

grafik

Hope you agree?

tinevez:
Go for it!

stefanhahmann:
> Go for it!

@maarzt may I ask you to merge (I do not have the rights to do so)? If you have questions regarding the remaining conflict, you may contact me.

tinevez:
Merged! But I had to do it via a local branch I had rights on. 
This PR can be marked as merged and closed.

PR #245, opened 2023-06-09
Notify MastodonPlugins about a Mastodon project being closed

### LINKED ISSUE #237: Notify MastodonPlugins about a Mastodon project being closed
# Problem

Currently there is no way for a MastodonPlugin to get notified when the Mastodon project is closed. This means it cannot free resources, that are associated with the Mastodon project or do other operations like storing settings etc. that would usually be done when closing a project.

This is a problem for the Lineage Registration plugin in mastodon-tomancak. The plugin keeps a list of opened Mastodon projects. But a Mastodon project is never removed from that list because the plugin simply doesn't know if a project gets closed or not.

# Potential Solution (deprecated)

<s>
The `MastodonPlugin` interface has a method that is called whenever a new Mastodon project is opened:

https://github.com/mastodon-sc/mastodon/blob/a965612e7890ec624232fc69c7431badd65c18d3/src/main/java/org/mastodon/app/plugin/MastodonPlugin.java#L51

We could add a similar method:
```java
void closingAppPluginModel( final M pluginAppModel )`
```
that is called whenever the Mastodon project is closed.</s> 

# Preferred Solution

Add `closingListerners` to the `AppModel`. A plugin can listen to the `AppModel` and would be notified when the `AppModel` is closed.

maarzt:
Fixes #237

This PR does two things:
* Adds a CloseListener to the MamutAppModel, this allows a MamutPlugin to get notified when the Mastodon project is closed.
* Allows a MamutPlugin to retrieve the project folder when initializing the plugin.

Here is an example:
```java
@Plugin( type = MamutPlugin.class )
public class ExampleMastodonPlugin implements MamutPlugin
{
	@Override
	public void setAppPluginModel( MamutPluginAppModel appPluginModel )
	{
		// register a close listener
		appPluginModel.getAppModel().closeListeners().add( this::onClosed );
		
		// get the project folder
		File projectFolder = appPluginModel
				.getWindowManager()
				.getProjectManager()
				.getProject()
				.getProjectRoot();
	}

	private void onClosed()
	{
		// TODO: free resources
	}
}
```

### DIFF
--- a/src/main/java/org/mastodon/mamut/MamutAppModel.java
+++ b/src/main/java/org/mastodon/mamut/MamutAppModel.java
@@ -42,6 +42,7 @@ import org.mastodon.views.bdv.SharedBigDataViewerData;
 import org.mastodon.views.bdv.overlay.ui.RenderSettingsManager;
 import org.mastodon.views.grapher.display.style.DataDisplayStyleManager;
 import org.mastodon.views.trackscheme.display.style.TrackSchemeStyleManager;
+import org.scijava.listeners.Listeners;
 import org.scijava.ui.behaviour.KeyPressedManager;
 import org.scijava.ui.behaviour.util.Actions;
 
@@ -72,6 +73,8 @@ public class MamutAppModel extends MastodonAppModel< Model, Spot, Link >
 	private final int maxTimepoint;
 
 	private final BranchGraphSynchronizer branchGraphSync;
+
+	private final Listeners.List< CloseListener > closeListeners = new Listeners.List<>();
 
 	public MamutAppModel(
 			final Model model,
@@ -156,5 +159,17 @@ public class MamutAppModel extends MastodonAppModel< Model, Spot, Link >
 	public BranchGraphSynchronizer getBranchGraphSync()
 	{
 		return branchGraphSync;
+	}
+
+	public void close() {
+		closeListeners.list.forEach( CloseListener::close );
+	}
+
+	/**
+	 * Listeners that are notified when the Mastodon project is closed,
+	 * by for example closing the main window.
+	 */
+	public Listeners< CloseListener > closeListeners() {
+		return closeListeners;
 	}
 }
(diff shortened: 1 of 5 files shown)


tinevez:
It's good. I see that you plugged the listener into the branch views. The branch views have already a `close()` method that should be called when the main window is closed. It does not work in your cases?

src/main/java/org/mastodon/mamut/MamutAppModel.java
```diff
@@ -157,4 +160,16 @@ public BranchGraphSynchronizer getBranchGraphSync()
...
+
+	public void close() {
+		closeListeners.list.forEach( CloseListener::close );
+	}
+
+	/**
+	 * Listeners that are notified when the Mastodon project is closed,
+	 * by for example closing the main window.
+	 */
+	public Listeners< CloseListener > closeListeners() {
```
stefanhahmann, line 172: The method name may be misleading and could also be interpreted as in "close the listeners". The javadoc explains it, but nevertheless maybe `getCloseListeners()` would be less ambiguous?

src/main/java/org/mastodon/mamut/MamutAppModel.java
```diff
@@ -157,4 +160,16 @@ public BranchGraphSynchronizer getBranchGraphSync()
 	{
 		return branchGraphSync;
 	}
+
+	public void close() {
```
stefanhahmann, line 164: Could you please check, if you have the correct code style settings? The bracket for the method should be a new line.

stefanhahmann (changes requested):
The PR adds the ability that Plugins can add listeners that inform, when a project is closed. This is a well needed functionality.

I left some comments re code style and method naming.

src/main/java/org/mastodon/mamut/MamutAppModel.java
maarzt: I renamed the method to projectCloseListeners()

stefanhahmann (approved):
My comments have been resolved. All good now from my perspective.

tinevez (approved):

PR #247, opened 2023-07-05
Adds a new class TagSetUtils

stefanhahmann:
This PR adds a new class TagSetUtils and TagHelper

* The class contains some convenience methods to deal with tag sets
* This class is moved from: https://github.com/mastodon-sc/mastodon-tomancak/blob/master/src/main/java/org/mastodon/mamut/tomancak/lineage_registration/TagSetUtils.java
* New methods tagSpotAndOutgoingEdges(), tagSpotAndIncomingEdges, tagSpot, tagSpots, tagLinks are added
* A unit test testing tagSpotAndOutgoingEdges() and addNewTagSetToModel() is added
* The tagSetAsMap() method is removed
* A findTag() method, that is similar to findTagSet
* A new class TagHelper is added that makes tagging of spots and links easier


maarzt (commented):
Looks good to me

PR #249, opened 2023-07-17
Add rgbaValidColor() convenience utility methods to TagSetUtils 

xulman:
This is motivated by my never-knowing if the alpha channel represents opacity or transparency (and thus the value to obtain a full solid color should be 1.0 or 0.0). So I propose the three convenience methods that, besides setting the proper or user alpha, build/encode the color into the integer.


xulman:
sorry, wrong target branch.. too late.

PR #255, opened 2023-08-29
GraphML file importer

tinevez:
Mastodon_GraphML-importer

This plugin can import data coming from a GraphML file (http://graphml.graphdrawing.org/) generated by Python software using networkx graph.
    
It requires the properties with keys 'x', 'y', 'z' and 'frame' to be present and can handle supplemental properties that are imported as Mastodon features. The features values are saved and restored along with the data.

A simple dialog allows specifying from where the take pixel size; apparently the x, y, z position are saved in pixel coordinates. Right now the radius or covariance matrix for the ellipsoid are not imported. The user must specify a radius.


tinevez:
Tested with the new dev branch. Ready for beta 27.

Issue #217, opened 2022-12-05
Link Data table, source/target spot are flipped


drchrisch:
Dear mastodon stuff,

recently I observed that target and source spot ids appear in flipped order in the Link Data table. Surprisingly, this seems to occur randomly but I can reach the expected order (source, timepoint n -> target, timepoint n+1) when I repeatedly run "compute features".

![Link DataTable](https://user-images.githubusercontent.com/42037938/205683003-7c679b0f-3a6b-4bad-9a8e-88a38baec341.png)

My input data are a bunch of 3D image stacks from a dozen timepoints, but I guess this does not really matter. Could you have a look into this?

Many thanks
Christian

PR #257, opened 2023-09-01
Update LinkTargetIdFeature.java

stefanhahmann:
Resolves: https://github.com/mastodon-sc/mastodon/issues/217

Replace `HashSet `used as return value for the projections() method by `LinkedHashSet`, since the order of entry needs to be stable in process using this method.

### DIFF
--- a/src/main/java/org/mastodon/mamut/feature/LinkTargetIdFeature.java
+++ b/src/main/java/org/mastodon/mamut/feature/LinkTargetIdFeature.java
@@ -31,7 +31,7 @@ package org.mastodon.mamut.feature;
 import static org.mastodon.feature.FeatureProjectionKey.key;
 
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 import org.mastodon.feature.Dimension;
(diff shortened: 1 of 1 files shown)


tinevez (approved):

stefanhahmann:
@tinevez Since the order of implementations of the org.mastodon.feature.Feature#projections() method needs to be reproducible, do you think, it would be good to change the return value of the method from Set to LinkedHashSet? By this, errors like the one fixed with this PR could not happen again.

cf. https://github.com/mastodon-sc/mastodon/blob/a965612e7890ec624232fc69c7431badd65c18d3/src/main/java/org/mastodon/feature/Feature.java#L66

tinevez:
@stefanhahmann Actually the order in which projections are returned should *not* matter. 
I think there is a deeper problem, probably rooted in the Table headers

stefanhahmann:
> @stefanhahmann Actually the order in which projections are returned should _not_ matter. I think there is a deeper problem, probably rooted in the Table headers

I agree. When fixing this bug, I was also assuming that there is something in table headers which does not appropriately handle the existing projections and thus relies on some order. However, I did not dig it down.

tinevez:
I think that the issue is related to the fact that everytime we call the `projections()` method in this feature, it returns a **new** HashSet, with **new** instances of the feature projection classes. 
But in the table, when building the header, the `projections()` method is called _twice_, once for the header subtitle and once to store the projection values. I think it comes from that.

Your patch fix the issue because it imposes the order. But I will try to fix the table so that the header and values are built from the same _instance_ of the projection, so that we don't depend on order. 

But also, I think it is sometimes nicer to impose a column order, like you did :)

tinevez:
See https://github.com/mastodon-sc/mastodon/commit/dca0534c45ae8cb995fa328bbceaf2dfcf4c0b49

PR #258, opened 2023-10-12
Update pom.xml

stefanhahmann:
Update pom-scijava from 33.2.0 to 37.0.0


stefanhahmann:
Replaced by https://github.com/mastodon-sc/mastodon/pull/261

PR #261, opened 2023-10-12
Update pom.xml with pom-scijava 37.0.0

stefanhahmann:
* Updates the pom-scijava version from 35.1.0 to 37.0.0
* KeymapManager, MastodonDebugSettings, WorkaroundConstructor and WorkaroundRepresenter needed to be adapted as well, since the empty constructors of Representer and Constructor do no exist anymore

### DIFF
--- a/src/main/java/org/mastodon/io/yaml/WorkaroundConstructor.java
+++ b/src/main/java/org/mastodon/io/yaml/WorkaroundConstructor.java
@@ -31,15 +31,17 @@ package org.mastodon.io.yaml;
 import java.util.List;
 import java.util.Map;
 
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.nodes.MappingNode;
 import org.yaml.snakeyaml.nodes.SequenceNode;
 
 public class WorkaroundConstructor extends Constructor
 {
-	public WorkaroundConstructor( final Class< ? > theRoot )
+
+	public WorkaroundConstructor( LoaderOptions loadingConfig )
 	{
-		super( theRoot );
+		super( loadingConfig );
 	}
 
 	@Override
--- a/src/main/java/org/mastodon/ui/keymap/KeymapManager.java
+++ b/src/main/java/org/mastodon/ui/keymap/KeymapManager.java
@@ -46,6 +46,7 @@ import org.scijava.ui.behaviour.io.InputTriggerDescription;
 import org.scijava.ui.behaviour.io.InputTriggerDescriptionsBuilder;
 import org.scijava.ui.behaviour.io.yaml.YamlConfigIO;
 import org.yaml.snakeyaml.DumperOptions;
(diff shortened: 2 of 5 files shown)


src/main/java/org/mastodon/io/yaml/WorkaroundConstructor.java
```diff
@@ -31,15 +31,17 @@
...
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.nodes.MappingNode;
 import org.yaml.snakeyaml.nodes.SequenceNode;
 
 public class WorkaroundConstructor extends Constructor
 {
-	public WorkaroundConstructor( final Class< ? > theRoot )
+
+	public WorkaroundConstructor( LoaderOptions loadingConfig )
```
tpietzsch, line 42: There is a `Constructor` constructor which takes `theRoot` and `loadingConfig`. Probably better to use that?

src/main/java/org/mastodon/ui/keymap/KeymapManager.java
```diff
@@ -229,9 +230,9 @@ private static Yaml createYaml()
 	{
 		final DumperOptions dumperOptions = new DumperOptions();
 		dumperOptions.setDefaultFlowStyle( DumperOptions.FlowStyle.BLOCK );
-		final Representer representer = new Representer();
+		final Representer representer = new Representer( new DumperOptions() );
```
tpietzsch, line 233: There is a `dumperOptions` constructed 2 lines above. Why not use that instead of a new one?

src/main/java/org/mastodon/util/MastodonDebugSettings.java
```diff
@@ -143,9 +144,9 @@ private static Yaml createYaml()
 		{
 			final DumperOptions dumperOptions = new DumperOptions();
 			dumperOptions.setDefaultFlowStyle( DumperOptions.FlowStyle.BLOCK );
-			final Representer representer = new Representer();
+			final Representer representer = new Representer( new DumperOptions() );
```
tpietzsch, line 147: There is a dumperOptions constructed 2 lines above. Why not use that instead of a new one?

tinevez:
I thinks we could also solve this by merging into `dev` the `rework-app-model` branch.
The latter is a recent attempt to make API breaking changes fixing my faulty design choices when I was young. It contains fixes for this issue. For instance:
https://github.com/mastodon-sc/mastodon/blob/rework-app-model/src/main/java/org/mastodon/io/yaml/WorkaroundConstructor.java

Then once in dev we can update the pom parent.
What do you think?

src/main/java/org/mastodon/util/MastodonDebugSettings.java
stefanhahmann: @tpietzsch thanks for your comments. The suggestions are good. I closed this PR because of https://github.com/mastodon-sc/mastodon/pull/261#issuecomment-1761275957

Issue #269, opened 2023-11-01
Garbage Collection Test Failes


maarzt:
Running the unit tests currently fails under certain conditions: Ubuntu, Java 11, with enabled graphics environment.
```
Running org.mastodon.model.ModelGarbageCollectionTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.33 sec

Results :

Failed tests: 
  testIfMastodonIsGarbageCollectable(org.mastodon.mamut.GarbageCollectionTest)
```

maarzt:
I'm going to fix it.

PR #270, opened 2023-11-02
Fix Garbage Collection

maarzt:
The unit test in `GarbageCollectionTest` fails currently, when executed with an enabled graphics environement.
This is because `TagSetDialog`, `FeatureComputationDialog` and `PreferencesDialog` are no longer disposed properly since the latest changes to the app model.

This PR fixes the problem by adding the dialog's `dispose()` methods to the `ProjectModel.projectClosedListeners()`.

### DIFF
--- a/src/main/java/org/mastodon/mamut/WindowManager.java
+++ b/src/main/java/org/mastodon/mamut/WindowManager.java
@@ -222,14 +222,18 @@ public class WindowManager
 		settings.addPage( new FeatureColorModeConfigPage( "Settings > Feature Color Modes", featureColorModeManager,
 				featureProjectionsManager, "Spot", "Link" ) );
 		settings.pack();
+		appModel.projectClosedListeners().add( settings::dispose );
 
 		/*
 		 * Tag-set and feature computation dialogs
 		 */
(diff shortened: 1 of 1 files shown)


maarzt:
fixes #269

maarzt:
Oh, this is a pity. I didn't expect the test to fail on other platforms. I thought the Java Swing implementation on linux would be the most unforgiving in terms a memory leaks. But the test passes on my machine. Could you please double check if everything is setup correctly?

Is your working directory clean? Did you checkout the correct "fix-garbage-collection" branch. Please make sure to run `mvn clean` before running the tests. (I had the problem on machine: The test was failing after I switched to the `fix-garbage-collection` git branch. Calling `mvn clean` fixed the problem. So I hope that's the same on your machine.)

@stefanhahmann Could you test if the test passes on windows?

maarzt:
Ok, @stefanhahmann did run the GarbageCollectionTest on Windows. It passes.

@tinevez I hope the test will pass on your Mac after executing "mvn clean". Worst case, the test continues to fail. This would likely mean, that there is a platform specific memory leak on Mac. We could have a video call, and try to debug the problem together. I can show you how I use the Eclipse Memory Analyzer to find memory leaks.

PR #278, opened 2024-02-08
Add the panel that showing current timepoint and number of spots to bdv views

### LINKED ISSUE #277: Add timepoint and number of spots info to BDV View
# User story:

Since version Beta-27, there is a new info in all track scheme windows, that shows the current timepoint and the number of spots of this timepoint:

grafik

I would like to have that same info also in the BDV window of Mastodon: 

grafik

stefanhahmann:
This PR adds the the info of the current timepoint and the current number of spots to the BDV

grafik

Resolves #277


stefanhahmann:
Wow. Probably my quickest ever merged PR. ;)

PR #289, opened 2024-03-07
Add Unit Test to Thoroughly Tests Mastodon's Undo Redo Mechanism

maarzt:
There is a PR to fix a small problem in Mastodon's undo/redo mechanism: https://github.com/mastodon-sc/mastodon-collection/pull/16

The goal of this PR is to ensure that Mastodons very important undo/redo mechanism works properly. I there for added a unit test `ModelUndoRedoTest`. The unit test uses the `ModelUtils.dump(...)` method to convert the current status of the `Model` into a string. Converting the model status into string allows me to test if the model is correctly rolled back into a specific (previously) recorded state.

It was necessary to extend the `ModelUtils.dump(...)` method to also print the tag sets. (The undo redo mechanism handle tags as well.)

### DIFF
--- a/src/main/java/org/mastodon/mamut/model/ModelUtils.java
+++ b/src/main/java/org/mastodon/mamut/model/ModelUtils.java
@@ -35,6 +35,7 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 
 import org.mastodon.collection.ref.RefArrayList;
@@ -126,20 +127,29 @@ public class ModelUtils
 	 */
 	public static final String dump( final Model model, final long maxLines, final DumpFlags... options )
 	{
-		EnumSet< DumpFlags > optionsSet = EnumSet.noneOf( DumpFlags.class );
-		optionsSet.addAll( Arrays.asList( options ) );
-		final String spaceUnits = Optional.ofNullable( model.getSpaceUnits() ).orElse( "" );
-		final ModelGraph graph = model.getGraph();
+		Set< DumpFlags > optionsSet = EnumSet.copyOf( Arrays.asList( options ) );
 		final FeatureModel featureModel = model.getFeatureModel();
-		Spot ref = graph.vertexRef();
 		final List< FeatureSpec< ?, ? > > featureSpecs = new ArrayList<>( featureModel.getFeatureSpecs() );
 		featureSpecs.sort( Comparator.comparing( FeatureSpec::getKey ) );
 
 		final StringBuilder str = new StringBuilder();
 		if ( optionsSet.contains( DumpFlags.PRINT_HASH ) )
 			str.append( "Model " ).append( model ).append( "\n" );
-		str.append( "Spots:\n" );
 
+		str.append( "Spots:\n" );
+		addSpotsTable( model, maxLines, optionsSet, featureSpecs, str );
+
+		str.append( "Links:\n" );
+		addLinksTable( model, maxLines, optionsSet, featureSpecs, str );
+
+		return str.toString();
+	}
+
+	private static void addSpotsTable( Model model, long maxLines, Set< DumpFlags > optionsSet, List< FeatureSpec< ?, ? > > featureSpecs, StringBuilder str )
+	{
+		final String spaceUnits = Optional.ofNullable( model.getSpaceUnits() ).orElse( "" );
+		final ModelGraph graph = model.getGraph();
+		final FeatureModel featureModel = model.getFeatureModel();
 		TablePrinter< Spot > spotsTable = new TablePrinter<>();
 		spotsTable.defineColumn( 9, "Id", "", spot -> Integer.toString( spot.getInternalPoolIndex() ) );
 		spotsTable.defineColumn( 9, "Label", "", Spot::getLabel );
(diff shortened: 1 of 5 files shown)


maarzt:
The test currently fails because of a but in the Undo/Redo mechanism. That will be fixed by this PR https://github.com/mastodon-sc/mastodon-collection/pull/16

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
```diff
@@ -0,0 +1,276 @@
...
+	/**
+	 * This method makes various changes to the graph and the tag set model. The
+	 * changes are divided into several steps, and at each step, an undo point is
+	 * set. Then, the undo and redo methods are called to test if the model is
+	 * restored to the intended state.
+	 */
+	private static void testUndoRedo( Model model )
+	{
+		final List< String > states = new ArrayList<>();
+		makeVariousChangesAndRecordUndoPoints( model, states );
```
stefanhahmann, line 125: I find it confusing when reading code, where changes to the variables that are given to the method happen inside the method. I usually do not expect this to happen. 
Could this method maybe return a Pair<Model, List> instead? That would confuse me less.

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
```diff
@@ -0,0 +1,276 @@
...
+		testUndoRedo( model );
+	}
+
+	/**
+	 * This method makes various changes to the graph and the tag set model. The
+	 * changes are divided into several steps, and at each step, an undo point is
+	 * set. Then, the undo and redo methods are called to test if the model is
+	 * restored to the intended state.
+	 */
+	private static void testUndoRedo( Model model )
```
stefanhahmann, line 122:
```suggestion
	private static void testUndoRedo( final Model model )
```
It seems like you forgot the final modifiers for the variables. It is not strictly speaking required, but seems to be a convention in this project.

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
```diff
@@ -0,0 +1,276 @@
...
+	 */
+	@Test
+	public void testUndoRedoAfterModelImporter()
+	{
+		Model model = new Model();
+		// Fill the undo-redo history with many different entries:
+		makeVariousChangesAndRecordUndoPoints( model, new ArrayList<>() );
+		addFourSpotGraph( model );
+		// Reset the model by running the ModelImporter / Simulate an import operation:
+		new ModelImporter( model )
```
stefanhahmann, line 230: This is difficult to read and understand what is happening and why it is written in this way.

However, the constructor of ModelImporter is protected and should not be made public only for this test thus I do not see a simpler alternative to this.

stefanhahmann (changes requested):
The main contribution of this PR is a unit test that checks if the undo/redo function works correctly on the Model. This is very valuable! Some parts of the test need some time to understand them, but overall the test is well written.
Also, this PR improves the dumping methods of ModelUtils a bit, which is also a good contribution.

I made some suggestion mainly regarding code formatting and also a bit regarding code structuring.

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
maarzt: I introduced a UndoVerifier class to solve that problem.

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
maarzt: There is now a Importer subclass that avoids this problem.

src/test/java/org/mastodon/mamut/model/ModelUtilsTest.java
```diff
@@ -0,0 +1,105 @@
...
+	{
+		Model model = new Model();
+		ModelGraph graph = model.getGraph();
+		Spot a = graph.addVertex().init( 0, new double[] { 1, 2, 3 }, 1 );
+		Spot b = graph.addVertex().init( 1, new double[] { 1, 2, 3.2 }, 1 );
+		a.setLabel( "A" );
+		b.setLabel( "B" );
+		graph.addEdge( a, b ).init();
+		String actual = ModelUtils.dump( model );
+		String expexted = "Model " + model + "\n"
```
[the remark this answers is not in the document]
maarzt: done

src/test/java/org/mastodon/mamut/model/ModelUtilsTest.java
```diff
@@ -0,0 +1,105 @@
...
+				+ "-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------\n"
+				+ "        0          A       0        1.0        2.0        3.0                   0                   1               1             0        1.0        2.0        3.0            1.0\n"
+				+ "        1          B       1        1.0        2.0        3.2                   1                   0               1             1        1.0        2.0        3.2            1.0\n"
+				+ "Links:\n"
+				+ "       Id  Source Id  Target Id    Link delta T    Link displacement    Source spot id    Target spot id    Link velocity\n"
+				+ "                                                             (pixel)                                        (pixel/frame)\n"
+				+ "-------------------------------------------------------------------------------------------------------------------------\n"
+				+ "        0          0          1             1.0                  0.2               0.0               1.0              0.2\n";
+		assertEquals( expexted, actual );
```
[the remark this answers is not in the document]
maarzt: done

maarzt (commented):
I addressed all the reviewers comments

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
stefanhahmann: This is better now. Perhaps `UndoRedoModifier` would be a bit better suited as class name?

src/test/java/org/mastodon/mamut/model/ModelUndoRedoTest.java
stefanhahmann: Good idea. I like it.

src/main/java/org/mastodon/mamut/model/ModelUtils.java
```diff
@@ -65,231 +74,233 @@ public static final String dump( final Model model )
...
+		if ( projection.isSet( t ) )
+			if ( projection instanceof IntFeatureProjection )
+				return String.format( "%" + width + "d", ( int ) projection.value( t ) );
+			else
+				return String.format( "%" + width + ".1f", projection.value( t ) );
+		else
+			return String.format( "%" + width + "s", "unset" );
+	}
+
+	private static class TablePrinter< T >
```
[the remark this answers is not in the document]
stefanhahmann: Okay.

src/main/java/org/mastodon/mamut/model/ModelUtils.java
```diff
@@ -65,231 +74,233 @@ public static final String dump( final Model model )
...
-	 * 
-	 * @param model
-	 *            the model.
-	 * @param maxLines
-	 *            the max number of rows to print in the two tables.
-	 * @return a String representation of the model.
+	 * Which information to include in the table can be changed using
+	 * {@link DumpFlags}.
+	 */
+	public static final String dump( final Model model, final DumpFlags... options )
```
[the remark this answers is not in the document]
stefanhahmann: I see. Thanks for the explanation. The link is an interesting read. There are pros and cons regarding this that I was not aware of.

stefanhahmann (approved):
Apart from a minor suggestion to rename UndoVerifier to UndoRedoVerifier, which I leave up to your decision, I have no further suggestions to this PR.

PR #302, opened 2024-04-05
Add new method to tag set utils

stefanhahmann:
This Pull Request adds some convenience methods to TreeUtils and TagSetUtils


src/main/java/org/mastodon/util/TreeUtils.java
tinevez, line 1: The method create a `ref` that the caller cannot release. Can you make the method signature to include a `ref` provided by the caller?

```java
public static Spot getFirstSpot( final Model model, final BranchSpot branchSpot, Spot ref )
```

tinevez (changes requested):
1 method signature change requested

tinevez (approved):

src/main/java/org/mastodon/util/TreeUtils.java
stefanhahmann: I made a change as requested. The whole ref thing still confuses me sometimes.
I thought that I had given credit to this by releasing inside the method:

```
                Spot ref = model.getGraph().vertexRef();
		Spot first = model.getBranchGraph().getFirstLinkedVertex( branchSpot, ref );
		model.getGraph().releaseRef( ref );
```

But maybe this was not the way this is intended to be used.

PR #306, opened 2024-05-23
Add svg and png export to trackscheme panels

stefanhahmann:
I recently added the functionality to export the content of the lineage classification dendrogram to SVG/PNG files, cf.: https://github.com/mastodon-sc/mastodon-deep-lineage/releases/tag/mastodon-deep-lineage-0.2.0

After this was released, I was asked by a user, if this could be added as a functionality to the track scheme windows as well.

This PR adds exactly this functionality to all TrackScheme Panels and the BDV Panel. It also adds a generic ExportUtils class that can be re-used in all plugin projects to achieve similar behavior in different components.

* UI view: 
grafik
* Resulting SVG: [trackscheme.zip](https://github.com/mastodon-sc/mastodon/files/15418057/trackscheme.zip)
* Resulting PNG: trackscheme

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,12 @@
 		<dependency>
 			<groupId>io.humble</groupId>
 			<artifactId>humble-video-all</artifactId>
-		</dependency>		
+		</dependency>
+		<!-- export of panels to svg -->
+		<dependency>
+			<groupId>org.jfree</groupId>
+			<artifactId>jfreesvg</artifactId>
+		</dependency>
 
 		<!-- Test dependencies -->
 		<dependency>
--- a/src/main/java/org/mastodon/mamut/views/bdv/MamutViewBdv.java
+++ b/src/main/java/org/mastodon/mamut/views/bdv/MamutViewBdv.java
@@ -63,6 +63,7 @@ import org.mastodon.model.FocusModel;
 import org.mastodon.model.HighlightModel;
 import org.mastodon.model.NavigationHandler;
 import org.mastodon.model.SelectionModel;
+import org.mastodon.ui.ExportViewActions;
 import org.mastodon.ui.FocusActions;
 import org.mastodon.ui.HighlightBehaviours;
 import org.mastodon.ui.SelectionActions;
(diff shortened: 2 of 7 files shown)


tinevez:
This is great! And it looks much simpler that the batik export in TrackMate's TrackScheme. 

If possible I would like to keep the right-click button for interactions with the data on the panel, as in the other views. Could you put the command in the JMenu?

stefanhahmann:
> This is great! And it looks much simpler that the batik export in TrackMate's TrackScheme.

Please do not overestimate the capabilities of this. This function only paints the contents of the current view to the SVG/PNG file, i.e. it will not be complete, if the trackscheme as seen on the screen is not completely inside the window.

Thus, it would be more honest, if the entry would says "Export current TrackScheme view to SVG/PNG".

Would you still like to have it?

> 
> If possible I would like to keep the right-click button for interactions with the data on the panel, as in the other views. Could you put the command in the JMenu?

I was also thinking about this. I decided for the right click, since this would offer the highest re-usability with the lineage classification dialog, where I also have it on the right click. However, putting it in the menu should not be too much effort. Would you see it in the `File`, `Window `or `View` menu?

tinevez:
> Please do not overestimate the capabilities of this. This function only paints the contents of the current view to the SVG/PNG file, i.e. it will not be complete, if the trackscheme as seen on the screen is not completely inside the window.> 
> Thus, it would be more honest, if the entry would say "Export current TrackScheme view to SVG/PNG".
> Would you still like to have it?

Yes.
What about `Export current view to SVG/PNG` ?

> I was also thinking about this. I decided for the right click, since this would offer the highest re-usability with the lineage classification dialog, where I also have it on the right click. However, putting it in the menu should not be too much effort. Would you see it in the `File`, `Window `or `View` menu?

I would say `File`, after a separator.

stefanhahmann:
> Yes. What about `Export current view to SVG/PNG` ?

Agreed.

> I would say `File`, after a separator.

It's now in the `File` menu of the TrackSchemePanels and the BDV View.

tinevez:
Maybe there is a way to add it to all views automatically. Can I have a go at it?

stefanhahmann:
> Maybe there is a way to add it to all views automatically. Can I have a go at it?

Yes, sure.

I was not sure, if it makes sense to have it in all views (e.g. table view?) and if it works well in all views (e.g. grapher -> svg might be problematic). Otherwise, I would have tried as well.

tinevez:
I made some minor changes to the PR:
- the constraint on the component that is painted is relaxed, so that we don't export the scroll bars.
- the actions are added to the grapher views. It really looks good!
And you are right: it does not make sense to add this to more views.
Are you ok as it is? Can I merge?

tinevez (approved):

stefanhahmann:
> I made some minor changes to the PR:
> 
>     * the constraint on the component that is painted is relaxed, so that we don't export the scroll bars.
> 
>     * the actions are added to the grapher views. It really looks good!
>       And you are right: it does not make sense to add this to more views.
>       Are you ok as it is? Can I merge?

I am currently out of office until mid July. If you can wait until then, I would be happy to give it a closer look, test and give feedback if necessary. If it is more urgent for you, I would also be fine, if you merge it.

stefanhahmann:
>       Are you ok as it is? Can I merge?

The only change I made is adding a separator line in the File menu of the grapher view before the exports. It is more consistent now.

PR #312, opened 2024-06-17
Prevent TrackScheme From Freezing When Removing A Root Spot

maarzt:
This PR fixes a bug that causes the TrackScheme window to freeze after the following two steps:
1. Use "View > Show Only Selected Tracks"
2. Delete one of the visible root nodes

Here is a screen cast, demonstrating how to reproduce the problem:

trackscheme-roots-bug

The bug is triggered because the `DefaultRootsModel` and `BranchTrackSchemeRootsModel` keep a list of roots to be shown in the TrackScheme view. If one of the roots is removed from the `ModelGraph`, it becomes invalid, but the invalid entry still remains in the list of roots. The exception is triggered when the invalid object is processed by the `PainterThread`.

This PR fixes the bug, by removing the invalid entry from the list of roots as soon as the corresponding Spot is removed from the `ModelGraph`.


maarzt:
@tinevez Disclaimer: The bug that is fixed in this PR was introduced by me. So I'm the one to blame here.

tinevez (approved):

PR #313, opened 2024-06-18
Add a new unit test for project loading and closing

stefanhahmann:
This PR adds a new unit test that tests, if projects can be loaded and closed without causing memory leaks / OutOfMemoryException


stefanhahmann:
I created the test, because I was interested in some specific behavior while loading/closing project. So, there it was not related to any bug or issue on Mastodon core. However, I think, it would be useful for the future, thus I would like to keep it.

PR #320, opened 2024-07-30
Reinstate the specific create view methods in WindowManager.

tinevez:
More verbose but more intuitive.


xulman:
this is far more intuitive!
I'm all for having them in the API
(despite they duplicate code in a way)

thank you @tinevez for adding them

PR #324, opened 2024-09-12
Introduce new method areListenersPaused() to interface SelectionModel

### LINKED ISSUE #322: Selection in BranchSpot Table is slow
# User Story

In a dataset with ~4 Million Spots and ~100.000 BranchSpots, I am trying the following things:

## Spots

1. Open Spot Table
2. Click on a row on the very top
3. Press `Shift` + `End` intending to select all spots
4. Selection is done rather quick

##  BranchSpots

1. Open BranchSpot Table
2. Click on a row on the very top
3. Press `Shift` + `End` intending to select all BranchSpots
4. Selection takes very long (multiple minutes if not even hours)

The Performance of selection in the BranchSpot table is too slow. Can it be improved?

stefanhahmann:
This PR introduces a new method `areListenersPaused()` to the interface `SelectionModel`

* The method `areListenersPaused()` is used in `BranchGraphSelectionAdapter` to find out, if listeners have been paused outside of this method and only in case they are not perform
  * pause listeners
  * select the branch vertex
  * resume listeners
* In case they are already paused, then they are neither paused again nor are they resumed after selecting the branch vertex
* Without this change a selection of many branch spots at the same time caused that resumeListeners() was called with high frequency resulting in low performance, when selecting many branch spots at the same time

Resolves #322

### DIFF
--- a/src/main/java/org/mastodon/adapter/SelectionModelAdapter.java
+++ b/src/main/java/org/mastodon/adapter/SelectionModelAdapter.java
@@ -162,4 +162,10 @@ public class SelectionModelAdapter< V extends Vertex< E >, E extends Edge< V >, 
 	{
 		selection.pauseListeners();
 	}
+
+	@Override
+	public boolean areListenersPaused()
+	{
+		return selection.areListenersPaused();
+	}
 }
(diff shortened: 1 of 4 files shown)


tinevez (approved):
Great finding and solution. 
Would you care to add some javadocs to the 3 listener methods in the `SelectionListener` interface?

stefanhahmann:
> Great finding and solution. Would you care to add some javadocs to the 3 listener methods in the `SelectionListener` interface?

Good idea. I tried my best to describe these 3 methods. Hopefully their function will be clearer with this in the future.

PR #329, opened 2024-10-08
Facilitate Custom File Formats - Required For Mastodon-Git Project

maarzt:
The ["Mastodon-Git"](mastodon-sc/mastodon-git) plugin aims to efficiently store Mastodon datasets inside a git repository. Unfortunately git performs very poorly when processing a history of large binary files as produced by the default Mastodon file format. Mastodon-Git therefor implements a customized Mastodon file format that can very efficiently be processed with git.

Two classes in the Mastodon core need to be public in order for mastodon-git to implement this custom file format that can still benefit from mastodon core features:
* fast serialization of `Spot` and `Link` into byte arrays (class `ModelSerializer`)
* and serialization of the GUI state into an XML (class `MamutViewStateXMLSerialization`)

(Mastodon-git currently has its own copies of the `ProjectLoader` and `ProjectSaver` classes. In the long term it would be amazing if we could add a plugin API into mastodon core. That supports the implementation of custom file formats. Or at least make the `ProjectLoader` and `-Saver` more reusable. However in the short term it would be great if the `ModelSerializer` and `MamutViewStateXMLSerialization` were public. This would allow us the further experiment with the mastodon-git file format, without the need to use a customized mastodon.jar file.)

### DIFF
--- a/src/main/java/org/mastodon/mamut/io/MamutViewStateXMLSerialization.java
+++ b/src/main/java/org/mastodon/mamut/io/MamutViewStateXMLSerialization.java
@@ -72,7 +72,7 @@ import net.imglib2.realtransform.AffineGet;
  * Utility class that can transform a GUI state
  * <code>Map< String, Object ></code> to XML and vice versa.
  */
-class MamutViewStateXMLSerialization
+public class MamutViewStateXMLSerialization
 {
 
 	private static final String WINDOW_TAG = "Window";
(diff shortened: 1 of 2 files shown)


tinevez:
Hello. Let's do this.
Your comments are very helpful. Maybe we should keep track of them in a github project if we revise the IO. 
I feel like Mastodon could be very useful as a viewer / editor for other tracking tools. A good extensible IO will be required if we are to go this way.

Issue #309, opened 2024-06-04
Provide textual way to search and run commands in Mastodon


stefanhahmann:
# User Story

In Fiji the, there is a search box that can be used to search and run FIJI commands.

![grafik](https://github.com/mastodon-sc/mastodon/assets/10515534/b3bb043c-93e1-4731-ae65-2d4908a99bdc)

This provides a convenient way to find and access functions for users, which does not require to search within the menus. It would be cool to have a similar mechanism also within Mastodon, since there are also a lot of functions in the core and also in the plugins, which are difficult to explore.

I could imagine three options how to provide this function:

1. Add a search box to Mastodon windows in the lower right corner just as in Fiji to search Mastodon and run Mastodon commands: 
![grafik](https://github.com/mastodon-sc/mastodon/assets/10515534/c0d41a4c-b420-4a3b-8481-1c338143056c)
2. Add a search box to the main Mastodon panel: 
![grafik](https://github.com/mastodon-sc/mastodon/assets/10515534/bda0cbb7-14c4-4842-88a0-73957f0f937b)
3. Add a new "Test" button in the Keymap settings that would run the currently filtered/selected command 
![grafik](https://github.com/mastodon-sc/mastodon/assets/10515534/64216c7f-b1dc-4917-b5f3-f0e7716b6466)

stefanhahmann:
I think this branch goes very much into the direction to what was asked by the user who approached me with this idea.

This would be my feedback after testing:
* It would be nice, if there was an alternative way to access the command finder. With only the `CTRL + Shift + F` shortcut, it would be a bit hidden. I think, if there was some additional visual search bar, users would benefit even more from the existence of the Mastodon command finder
* The list of available commands seems incomplete:
  * E.g., if opened from BDV, I cannot set the 2d drag rotate in the command finder, while in the dialog `File > Preferences > KeyMap`, I can: 
![grafik](https://github.com/mastodon-sc/mastodon/assets/10515534/b8cd2679-5f04-4faf-bed7-d3d701c38e86)
  * It is unclear to me how commands from plugins could be added to the command finder
* The command finder shows duplicates, because for some commands multiple short cuts are defined. For such situations, it would be nice, if only one row would be used that shows all shorts cuts. Cf. e.g. this example from IntelliJ: 
![grafik](https://github.com/mastodon-sc/mastodon/assets/10515534/921da4ce-8092-4c48-9e71-8034de1f8d0c)
[...]

PR #331, opened 2024-10-09
A command finder for Mastodon windows.

tinevez:
Based on a stripped down version of org.scijava.ui.behaviour.io.gui.VisualEditorPanel


stefanhahmann:
Resolves #309

PR #332, opened 2024-10-09
Rework import / export menus

tinevez:
There are now two submenus in the `File` main menu that are made for plugins that import or export data in the current mastodon session. 
With this PR, the `Plugins` menu of a 'naked' mastodon application (with no other artifact present) is empty and not shown.

Screenshot 2024-10-09 at 15 47 38


tinevez:
> The actual changes look good to me.
> 
> However, the CI pipeline does not run through. Can adapt the javadoc of the class MamutViewStateXMLSerialization so that it does?
> 

I take care of this.

PR #344, opened 2024-11-07
More complete restore of grapher gui state

stefanhahmann:
This PR makes the following changes:

* Only zoom out to full data extent in Grapher after Button "Plot" has been pressed
* In cases, when the grapher is redrawn after changes, the screen transform will stay the same. This includes cases, when the Grapher is redrawn after opening a project
* Add serialization of the GUI state of the grapher views. The following things are restored additionally the window size+position:
  * Selected X-Axis Feature + Projection
  * Selected Y-Axis Feature + Projection
  * The ScreenTransform that was used during saving the project
  * Whether or not to connect edges

The context chosen by the user can unfortunately not be restored, since the BDV View Windows have no unique ID. However, this is consistent to other views (e.g. TrackScheme), where the context can also not be restored.

grafik


tinevez (approved):
Tested live, and checked syntax and conventions.

tinevez:
Awesome work.

PR #353, opened 2024-11-14
Support more than 23 million spots in Mastodon

maarzt:
Using the MultiArrayMemPool rather than SingleArrayMemPool eliminates the upper bound for the number of Spots in Mastodon. Without this change Mastodon is limited to approx. 23.3 million spots. Reason for that is that the SingleArrayMemPool limits the available space for spots to 2 GB. MultiArrayMemPool removes this limitation.

### DIFF
--- a/src/main/java/org/mastodon/mamut/model/SpotPool.java
+++ b/src/main/java/org/mastodon/mamut/model/SpotPool.java
@@ -31,7 +31,7 @@ package org.mastodon.mamut.model;
 import org.mastodon.model.AbstractSpotPool;
 import org.mastodon.pool.ByteMappedElement;
 import org.mastodon.pool.ByteMappedElementArray;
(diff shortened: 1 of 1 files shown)


maarzt:
The are other similar pools in the source code, that probably would need to be changed to avoid similar problems these are at least:
* LinkPool
* DataGraph
* TrackSchemeGraph

tinevez:
Can we slow down on this?
I am not sure we want to do this change yet before having a use case. And particularly before the preprint.

stefanhahmann:
> Can we slow down on this? I am not sure we want to do this change yet before having a use case. And particularly before the preprint.

I would not mind slowing down on this.

However, this PR was in fact motivated while writing the manuscript. Johannes wants to make a point about the max. data Mastodon can handle. The number he can write depends on this PR.

PR #361, opened 2024-12-18
Highlight preferably spot at current timepoint

stefanhahmann:
This PR suggests the following change:

In case the highlighting is triggered from a branch view, try to highlight the spot in the branch at the current timepoint instead of the always the last spot in graph

* In coupled views this allows to show the highlighted spot in the BDV

Highlighting before this PR:
current highlighting

Highlight after this PR:
adapted highlighting

### DIFF
--- a/src/main/java/org/mastodon/model/branch/BranchGraphHighlightAdapter.java
+++ b/src/main/java/org/mastodon/model/branch/BranchGraphHighlightAdapter.java
@@ -28,6 +28,8 @@
  */
 package org.mastodon.model.branch;
 
+import java.util.Iterator;
+
 import org.mastodon.graph.Edge;
 import org.mastodon.graph.GraphIdBimap;
 import org.mastodon.graph.ReadOnlyGraph;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mastodon/model/branch/BranchGraphHighlightAdapter.java
```diff
@@ -48,28 +52,54 @@ public class BranchGraphHighlightAdapter<
...
-		final V vRef = graph.vertexRef();
-		highlight.highlightVertex( branchGraph.getLastLinkedVertex( vertex, vRef ) );
-		graph.releaseRef( vRef );
+		final V spotRef = graph.vertexRef();
+		Iterator< V > vertices = branchGraph.vertexBranchIterator( branchVertex );
+		boolean highlighted = false;
+		try
+		{
+			// prefer to highlight the spot at the current timepoint
+			while ( vertices.hasNext() )
```
tinevez, line 84: This might have too high of a cost on performance. I am afraid that it would compromise the responsiveness.

tinevez: > Hm, I admit that for the showing the last spot (as in the current implementation), there is a direct shortcut with a map holding the first/last spot for each branchspot.

Ah? Could you clarify this point?

>On the other side: typical tracking projects rarely have branchspots longer than 1000 spots. And for every highlighting, it would only be required to iterate through the spots of only the one branchspot, which is highlighted. Thus, normally it would not be too much of iterations required.

Yes you are right. Let's make it work.

PR #362, opened 2024-12-19
Remove context chooser from MamutBranchViewGrapher

stefanhahmann:
This PR removes the context chooser from the MamutBranchViewGrapher.

With the current implementation, it produces errors, when contexts other the "full graph" are chose.

Similarly to the MamutBranchViewTrackScheme, the context chooser can currently not really be used in MamutBranchView(s), since the context chooser / the context provider can provide spots, but in the branch view branch spots are required and translation between them is with the current architecture not possible


tinevez:
All good.
Maybe remove the remaining empty 'from context' label:
Screenshot 2024-12-20 at 11 05 18

tinevez:
Ok maybe I don't get this part right. When I test this PR, it seems that the 'Items to plot' selection is not respected on the branch grapher.

stefanhahmann:
> All good. Maybe remove the remaining empty 'from context' label:

I decided to keep this radio button, but rename to the only possible context, i.e. "full graph" in the grapher branch view:

grafik

Without this radio button, there would not be a possibility for the user to go back to "full graph", after e.g. "selection" or "track of selection had been chosen".

stefanhahmann:
> Ok maybe I don't get this part right. When I test this PR, it seems that the 'Items to plot' selection is not respected on the branch grapher.

Thanks for testing and spotting this. My PR indeed destroyed that the selection of 'Items to plot' was respected. I force pushed a change and now this selection is respected again.

stefanhahmann:
@tinevez I addressed your comments. Can you have a look on this PR again?

tinevez (approved):

PR #366, opened 2025-03-26
Change the default visibility of the color bar to true

stefanhahmann:
By chance I recently found out that a feature that I was always missing does actually exist: a legend for the Grapher views

grafik

Since, I think this feature is so super important and perhaps I am not the only one for whom it is too well hidden how to activate it (it is actually not hidden at all, but Mastodon has grown so powerful with so many options that for newbees, it might be hard to find the relevant stuff), I would like to suggest to change the default such that it is activated, which is done in this PR.


tinevez:
I intead appreciate if the scale bar would be off by default...

PR #368, opened 2025-04-09
Access grapher context chooser

stefanhahmann:
This PR exposes some methods related to the grapher so that they can be re-used

* The search panel is added to the grapher branch view

Related to and preparing for: https://github.com/mastodon-sc/mastodon-grapher-opengl/issues/4


src/main/java/org/mastodon/views/grapher/display/GrapherSidePanel.java
```diff
@@ -430,6 +435,11 @@ public FeatureProjectionSpec getFeatureProjectionSpec( final String projectionKe
 		return null;
 	}
 
+	public ContextChooser< Spot > getContextChooser()
+	{
+		return Cast.unchecked( this.contextChooser );
+	}
```
tinevez, line 443: All good. Since we actually have a `getContextChooser()` method, we can implement the 

```
org.mastodon.views.context.HasContextChooser< T >
```
interface.

tinevez (approved):

src/main/java/org/mastodon/views/grapher/display/GrapherSidePanel.java
tinevez: Well actually it is not critical, since it is not a top-level view class. What do you think?

stefanhahmann: Indeed, it is not critical.
But, actually a good idea to it. I added a commit, which does it.

PR #369, opened 2025-04-10
Let help button open the online docs in LauncherGUI

stefanhahmann:
I accidentally discovered that the help / question mark button in the launcher GUI has no function.

I am not sure what the initial intention for this button was, but this PR at least lets this button open the online docs as well.

grafik

### DIFF
--- a/src/main/java/org/mastodon/mamut/launcher/LauncherGUI.java
+++ b/src/main/java/org/mastodon/mamut/launcher/LauncherGUI.java
@@ -198,6 +198,8 @@ class LauncherGUI extends JPanel
 		c.gridwidth = 1;
 		c.gridy++;
 		btnHelp = new JButton( HELP_ICON_MEDIUM );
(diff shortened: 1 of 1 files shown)


tinevez:
Euh yes it has a function.
When clicked it shows the first page (the one displayed on your screenshot) where there is a clickable link to the doc.

stefanhahmann:
> Euh yes it has a function. When clicked it shows the first page (the one displayed on your screenshot) where there is a clickable link to the doc.

Indeed. 
I was pressing the button, while I actually was on the first page and in that case nothing happened, because the Help page was already shown. I then had a look in `LauncherGUI`, where I did not see any action listener on the button (because the ActionListener is defined in another class `MastodonLauncher`).. 🤦‍♂️

All good then, except from maybe other _dumbest user imaginable_ who would perhaps also wonder what the button does, when they press it accidentally, while they are already on the first page. But they will not wonder for too long - so no need to take any action here.

PR #370, opened 2025-04-23
Cosmetic change in MainWindow

stefanhahmann:
The MainWindow layout appears a bit ugly in Linux environments. I suggest to make the window a bit wider to overcome this issue.

Before (Linux): 

grafik

After (Linux):

grafik

Before (Windows):

grafik

After (Windows):

grafik

### DIFF
--- a/src/main/java/org/mastodon/mamut/MainWindow.java
+++ b/src/main/java/org/mastodon/mamut/MainWindow.java
@@ -111,7 +111,7 @@ public class MainWindow extends JFrame
 
 		// Main Panel
 		final JPanel buttonsPanel = new JPanel();
-		buttonsPanel.setLayout( new MigLayout( "wrap 2", "[150px!]10[150px!]" ) );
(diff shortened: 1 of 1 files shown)


tinevez:
Any chance we could use a smaller font size on linux instead? 
My reasoning is that this window just has buttons. The text is nice but not so informative, so making room for it is not super great. 
Also I would like to keep some width for later, as I would like to attach a side panel for logging.

stefanhahmann:
> Any chance we could use a smaller font size on linux instead? My reasoning is that this window just has buttons. The text is nice but not so informative, so making room for it is not super great. Also I would like to keep some width for later, as I would like to attach a side panel for logging.

I added a commit that changes the default font for unix/linux system. Since, I did not want to make that font smaller, I still had to increase the size of the MainWindow a bit (170), but less than in the first commit (where it was 190):

After PR (Linux):
grafik

After PR (Windows):
grafik

PR #372, opened 2025-07-11
Fix Bdv too frequent navigation deadlock

### LINKED ISSUE #371: Mastodon freezes when navigating too quickly with connected BDV
# How to reproduce?

1. Unzip and open [lineage_motifs.zip](https://github.com/user-attachments/files/21181849/lineage_motifs.zip)
2. Make sure table view and Big Data Viewer view are in connected views (e.g. Lock group 1)
3. Navigate very quickly with the arrow keys on the keyboard through the table (including maybe back and forth). And also in particular including switching between spots that are spatially distant from each other so that the connected BDV has to perform some subsequent navigation / animation
4. The Big Data Viewer window, and thus whole Mastodon, gets stuck after a while doing the steps explained in 3.

Image

stefanhahmann:
This PR contains a strategy to avoid to frequent navigation requests to the Big Data Viewer, which cause Mastodon to get stuck in certain sitations.

It is not the actual bug fix (that probably needs to be done in Big Data Viewer), but a work around to avoid the Big Data Viewer to get into this bug.

Fixes #371 

### Improvements to navigation frequency control:
* Added a `lastNavigationTime` field to track the timestamp of the last navigation request and prevent navigation requests that occur too frequently. This logic was implemented in both `navigateToVertex` and `navigateToEdge` methods. [[1]](diffhunk://#diff-480a1a258e537d21add1b0c2c86d7c7d773c621d8dda605e42cc1fad5ccc79d5R49-R52) [[2]](diffhunk://#diff-480a1a258e537d21add1b0c2c86d7c7d773c621d8dda605e42cc1fad5ccc79d5R89-R93) [[3]](diffhunk://#diff-480a1a258e537d21add1b0c2c86d7c7d773c621d8dda605e42cc1fad5ccc79d5L93-R116) [[4]](diffhunk://#diff-480a1a258e537d21add1b0c2c86d7c7d773c621d8dda605e42cc1fad5ccc79d5L114-R128)

### DIFF
--- a/src/main/java/org/mastodon/views/bdv/overlay/OverlayNavigation.java
+++ b/src/main/java/org/mastodon/views/bdv/overlay/OverlayNavigation.java
@@ -46,6 +46,10 @@ public class OverlayNavigation< V extends OverlayVertex< V, E >, E extends Overl
 
 	private NavigationBehaviour< V, E > navigationBehaviour;
 
+	private long lastNavigationTime = 0;
+
+	private static final int ANIMATION_DURATION = 300;
+
 	public OverlayNavigation(
 			final AbstractViewerPanel panel,
 			final OverlayGraph< V, E > graph )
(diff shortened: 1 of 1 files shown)


tinevez:
This is an horrid bug and a sensitive part of the code.
I would like to have the review of @tpietzsch here instead of mine.

tpietzsch:
I'll look at it during the celltrackathon

stefanhahmann:
Just a reminder, since users seem to be running into this bug quite frequently, when heavily using mastodon..

